PR #68, opened 2016-10-17
Allow replacement of singular value using a value filter if the value…

nprager:
… filter uses the special-case "value" path.


nprager:
This might be slightly non-standard SCIM behavior but it fixes the issue and seems pretty harmless.    If you know of a better place to fix it I am open to alternatives.

PR #69, opened 2017-01-11
DS-16678: remove message text and subject from validation request obj…

nprager:
…ects.


digitalperk:
Looks good to me as well!

PR #74, opened 2017-03-13
Update GitHub URLs to use 'pingidentity' organization instead of 'UnboundID'

braveulysses:
All public UnboundID repositories were recently transferred to the Ping Identity organization. While existing GitHub URLs should continue to work, it makes sense to use the canonical URLs going forward. This rather boring PR updates the GitHub URLs in the READMEs and POMs to use the 'pingidentity' organization instead of 'UnboundID'.

JiraIssue: DP-281

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,9 +37,9 @@ For clients using UnboundID-specific features:
 </dependency>
 ```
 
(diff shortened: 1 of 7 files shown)


digitalperk:
Looks good!

PR #82, opened 2017-08-11
Allow extending Meta to support extra attributes.

### LINKED ISSUE #81: Ignoring unknown fields?
Hey guys,

Question, we have under the 'meta' section some extra fields, when retrieving a resource we get:

```
java.lang.RuntimeException: com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "foo" (class com.unboundid.scim2.common.types.Meta), not marked as ignorable (5 known properties: "resourceType", "location", "lastModified", "created", "version"])
```

We do have in our schema a description of these fields but I guess the library doesn't consult with the schema rather just try to deserialize the Meta.class object,

Is there any way to ignore unknown fields or make the retrieve be aware of them somehow?

Another option would be to extends the resources to fit to our own needs and override the 'Meta' object (in case you are willing to make the `Meta` class not a final one)

Thanks

asaf:
Fixes #81

### DIFF
--- a/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/types/Meta.java
+++ b/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/types/Meta.java
@@ -25,7 +25,7 @@ import java.util.Calendar;
 /**
  * Stores metadata about a SCIM object.
  */
-public final class Meta
+public class Meta
 {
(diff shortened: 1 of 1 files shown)


kqarryzada:
Support for this will be added in the upcoming 4.0.0 release.

PR #84, opened 2017-08-15
Possibility to write the empty Resource from ListResponseWriter

Vikjen:


kqarryzada:
Hi, it's not clear what this code change seeks to address. If this is still a problem, please open an issue with some background on the desired behavior.

PR #88, opened 2017-10-11
Fix issues involving binary attribute handling

braveulysses:
* GenericScimResource getter and setters now work correctly when the value is stored in a TextNode or a BinaryNode. Previously, calling (e.g.) GenericScimResource.getBinaryValue(...) could result in a NullPointerException.
* SchemaChecker no longer incorrectly reports an issue if the value is stored in a BinaryNode.

The changes in this PR were done by @digitalperk; I'm simply requesting that we accept these changes into master. Please see the internal issue DS-18091.

### DIFF
--- a/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/GenericScimResource.java
+++ b/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/GenericScimResource.java
@@ -17,16 +17,11 @@
 
 package com.unboundid.scim2.common;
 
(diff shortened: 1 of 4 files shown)


braveulysses:
About the failed Travis CI tests: They're failing when targeted to Oracle JDK 7. It looks like the container that Travis uses no longer supports the oraclejdk7 target, because Oracle JDK 7 is EOL'd. I'll simply remove that build target in a separate PR. The other JDKs continue to work.

Issue #87, opened 2017-09-06
ErrorResponse status should be string


asaf:
Hey,

According to SCIM v2 spec, the "status" field of an error message should contain the status code represented a string (don't ask me why),

Will you accept a patch that fixes this?

p.s: It makes sense to keep the status code in the `ErrorResponse` class as int but just serialize it back as string via `ToStringSerializer` Jackson serializer

Thanks.

digitalperk:
Thank you for identifying the issue! I've created an internal JIRA issue to track issue and get it fixed and tested. There is no estimated time for resolution as the issue needs to be reviewed and prioritized.

braveulysses:
Thanks again for reporting this, @asaf. This was fixed a year ago, but it looks like I forgot to close the issue.

mmoayyed:
This seems like it's still an issue:

```
Caused by: javax.ws.rs.ProcessingException: Error reading entity from input stream.
        at org.glassfish.jersey.message.internal.InboundMessageContext.readEntity(InboundMessageContext.java:889) ~[jersey-common-2.27.jar:?]
        at org.glassfish.jersey.message.internal.InboundMessageContext.readEntity(InboundMessageContext.java:808) ~[jersey-common-2.27.jar:?]
        at org.glassfish.jersey.client.ClientResponse.readEntity(ClientResponse.java:321) ~[jersey-client-2.27.jar:?]
        at org.glassfish.jersey.client.InboundJaxrsResponse$1.call(InboundJaxrsResponse.java:115) ~[jersey-client-2.27.jar:?]
        at org.glassfish.jersey.internal.Errors.process(Errors.java:316) ~[jersey-common-2.27.jar:?]
        at org.glassfish.jersey.internal.Errors.process(Errors.java:298) ~[jersey-common-2.27.jar:?]
        at org.glassfish.jersey.internal.Errors.process(Errors.java:229) ~[jersey-common-2.27.jar:?]
        at org.glassfish.jersey.process.internal.RequestScope.runInScope(RequestScope.java:389) ~[jersey-common-2.27.jar:?]
        at org.glassfish.jersey.client.InboundJaxrsResponse.runInScopeIfPossible(InboundJaxrsResponse.java:264) ~[jersey-client-2.27.jar:?]
        at org.glassfish.jersey.client.InboundJaxrsResponse.readEntity(InboundJaxrsResponse.java:112) ~[jersey-client-2.27.jar:?]
[...]

PR #90, opened 2017-10-18
Serialize ErrorResponse status to string

braveulysses:
Previously, the ErrorResponse status value was serialized to a JSON number, but [RFC 7644](https://tools.ietf.org/html/rfc7644#section-3.12) actually calls for the field to be represented as a string. A custom serializer was added so that the status field is now serialized as a string; a custom deserializer was added that allows an ErrorResponse to be deserialized regardless of whether the status field is a string or a number. The SDK's ErrorResponse Java API is unchanged.

This addresses Issue #87.

JiraIssue: DS-18095

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
-## v2.1.3 - 2018-xx-xx
+## v2.2.0 - 2018-xx-xx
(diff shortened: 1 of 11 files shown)


braveulysses:
Clients written using this version of the SCIM 2 SDK should be backwards-compatible with servers written using earlier versions of the SDK. But older SDK clients and non-SDK clients that expect status to be a JSON number will not be compatible with servers written with this version of the SDK. @digitalperk, should we rev the SDK version to 2.2 or even 3.0?

digitalperk:
It may be nice to add a serialization test as well. Otherwise, looks great!

PR #137, opened 2020-01-22
Prevent possibility of template variable injection in SCIM2 SDK POST search

coulbeck:
Prevent possibility of template variable injection in SCIM2 SDK POST search

JiraIssue: DS-41106
Product: ds
Product: proxy
Product: broker

See DS-41106

DS-41106


coulbeck:
UriBuilder methods always interpret path template variables and I could not find any good alternative to UriBuilder. So I chose to percent encode left and right brace characters for the string query parameters before passing to UriBuilder.

braveulysses:
Looks good to me, Andy. If it's not too much, it would be nice to add a simple unit test for the encodeTemplateNames() method to the SDK.

PR #142, opened 2020-02-19
performance improvement for consecutive client requests:

### LINKED ISSUE #141: Performance issue in scim2-sdk-client
I've noticed that the scim2-sdk-client does not completely consume result entity data from client operations.  The effect is that the underlying TCP stream is automatically closed after each operation.  A new TCP stream must be re-negotiated for a subsequent request.  This is all automatic under Jersey, but it causes a delay that is especially noticeable when any of these things are true:
- there are many client requests
- TLS is used to secure the connection
- there is significant latency to the target server.

This can be reproduced with this client code:

```java
        ScimService scimService = new ScimService(target);
        for (int i = 0; i < 100; i++) {
            UserResource newUser = new UserResource().setUserName("createLoopUser" + i);
            UserResource createdUser = scimService.createRequest("Users", newUser).
                    contentType(MediaType.APPLICATION_JSON).invoke();
            Assert.assertNotNull(createdUser);
        }
```

This code will complete without errors, but you can observe the unnecessary network activity by examining a wireshark trace of the TCP/IP network communications.

gclayburg:
create, replace, search

This fix allows the client to reuse the underlying TCP stream and avoid creating
a new TCP connection for each request.  The effect of this is
especially noticeable when the connection to the SCIM server is using TLS or has
 significant latency.

Ok, I see that ping identity does not accept pull requests, but I suspect that you will want this fix.  The performance impact is quite significant under the conditions listed in issue #141 

Notes:
- I have tested this with GET, PUT and POST operations.  I have not tried it with PATCH requests, but I suspect there is probably a similar issue there that is not addressed by this PR.
- After implementing the fix, I noticed that the test class EndpointTestCase will fail in unpredictable ways.  However, it seems to me that this is somehow an artifact of how the test is constructed and not an issue with the fix in this PR itself.  I can supply the test case I use to verify this if it helps.  

_Please be aware that Ping Identity does not accept third-party contributions at this time! Please see our [contribution guidelines](https://github.com/pingidentity/scim2/blob/master/CONTRIBUTING.md)._

…
fixes #141

### DIFF
--- a/scim2-sdk-client/src/main/java/com/unboundid/scim2/client/requests/CreateRequestBuilder.java
+++ b/scim2-sdk-client/src/main/java/com/unboundid/scim2/client/requests/CreateRequestBuilder.java
@@ -75,6 +75,11 @@ public final class CreateRequestBuilder<T extends ScimResource>
       if(response.getStatusInfo().getFamily() ==
           Response.Status.Family.SUCCESSFUL)
       {
+        response.bufferEntity();
+        /* fully read and close the input stream, allowing the buffered result
+           to be used while still keeping the TCP connection open for subsequent
+           requests
+         */
         return response.readEntity(cls);
       }
       else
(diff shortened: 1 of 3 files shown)


kqarryzada:
As stated in #141, we won't be adding changes related to this, since the use of connection pooling is a much stronger solution.

PR #144, opened 2020-05-18
Prevent template variable injection into resource location URL

lance-purple-unboundid:
JiraIssue: DS-40219

### DIFF
--- a/scim2-sdk-server/src/main/java/com/unboundid/scim2/server/providers/AuthenticatedSubjectAliasFilter.java
+++ b/scim2-sdk-server/src/main/java/com/unboundid/scim2/server/providers/AuthenticatedSubjectAliasFilter.java
@@ -33,9 +33,9 @@ import javax.ws.rs.core.SecurityContext;
 import javax.ws.rs.core.UriBuilder;
 import javax.ws.rs.ext.Provider;
 import java.io.IOException;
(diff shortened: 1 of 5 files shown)


coulbeck:
This looks good. I did a search for all usages of UriBuilder (I should have done that when I fixed DS-41106), and found one more possible trouble spot.

It's in AuthenticatedSubjectAliasFilter, lines 68-80. This is where we handle a request to /Me.

Could you also add a call to encodeTemplateNames on line 77 where we call UriBuilder#queryParam.

```
          UriBuilder newRequestUri =
              requestContext.getUriInfo().getBaseUriBuilder();
          newRequestUri.path(authSubjectPath +
              requestPath.substring(alias.length()));
          MultivaluedMap<String, String> queryParams =
              requestContext.getUriInfo().getQueryParameters();
          for (String key : queryParams.keySet())
          {
            List<String> values = queryParams.get(key);
            newRequestUri.queryParam(key, values.toArray()); <<<< Template Injection Problem??
          }

          requestContext.setRequestUri(newRequestUri.build());
```

lance-purple-unboundid:
Updated AuthenticatedSubjectAliasFilter to encode the queryParams, and also the authSubjectPath (which can produce the IllegalArgumentException if /Me resolves to a resource containing curly braces)

PR #158, opened 2021-06-26
Add a mechanism to optionally permit semicolons in filtered attribute…

lance-purple-unboundid:
… names.

This is needed by the Ping Directory REST API, in order to support filtering
and retrieval of SCIM resources by LDAP attribute names, which may include
semicolon-delimited attribute options (see RFC2251 section 4.1.5 for details).

JIRA issue: DS-37117


scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/Parser.java
```diff
@@ -161,6 +177,76 @@ public static Filter parseFilter(final String filterString)
...
+   *
+   * <pre>
+   *   try
+   *   {
+   *     Parser.addOptions(ParserOption.ALLOW_SEMICOLONS_IN_ATTRIBUTE_NAMES();
+   *     performWhateverProcessing();
+   *   }
+   *   finally
+   *   {
+   *     Parser.removeOptions(ParserOption.ALLOW_SEMICOLONS_IN_ATTRIBUTE_NAMES();
```
davidely-ping, line 195: The method should instead call setOptions with whatever the options were before addOptions was called. This guards against the ParserOption.ALLOW_SEMICOLONS_IN_ATTRIBUTE_NAMES option already being set.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/Parser.java
```diff
@@ -161,6 +177,76 @@ public static Filter parseFilter(final String filterString)
...
+   * Add parser options, in order to modify parser behavior within the current
+   * thread.
+   *
+   * <p>NOTE: SCIM server implementations are not guaranteed to support a given option.</p>
+   * <p>NOTE: these should be removed when no longer needed.</p>
+   *
+   * <pre>
+   *   try
+   *   {
+   *     Parser.addOptions(ParserOption.ALLOW_SEMICOLONS_IN_ATTRIBUTE_NAMES();
```
davidely-ping, line 190: There's a typo at the end of this that makes ALLOW_SEMICOLONS_IN_ATTRIBUTE_NAMES look like a method call.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/Parser.java
```diff
@@ -161,6 +177,76 @@ public static Filter parseFilter(final String filterString)
...
+   *   }
+   *   finally
+   *   {
+   *     Parser.removeOptions(ParserOption.ALLOW_SEMICOLONS_IN_ATTRIBUTE_NAMES();
+   *   }
+   * </pre>
+   *
+   * @param options  The parser options to be added.
+   */
+  public static void addOptions(final ParserOption... options)
```
davidely-ping, line 201: This method could return the original options, which would make the new code above a little simpler simpler since it wouldn't need to call getOptions first.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/Parser.java
```diff
@@ -161,6 +177,76 @@ public static Filter parseFilter(final String filterString)
...
+    Parser.threadLocalOptions.set(updatedOptions);
+  }
+
+  /**
+   * Remove parser options, in order to resume default parser behavior within
+   * the current thread.
+   *
+   * @param options  The parser options to be removed.
+   */
+  public static void removeOptions(final ParserOption... options)
```
davidely-ping, line 217: Can we take this method out and just use setOptions? I'm concerned that having it will make it easier to write incorrect code like in the example above where an option will always be removed even if it was already set.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/Parser.java
```diff
@@ -161,6 +177,76 @@ public static Filter parseFilter(final String filterString)
...
+    Set<ParserOption> updatedOptions = Parser.threadLocalOptions.get();
+    for (ParserOption option : options)
+    {
+      updatedOptions.remove(option);
+    }
+    Parser.threadLocalOptions.set(updatedOptions);
+  }
+
+  /**
+   * Get the current set of parser options.
```
davidely-ping, line 228: Like with the other methods, I'd mention that this is just for the current thread.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/Parser.java
```diff
@@ -161,6 +177,76 @@ public static Filter parseFilter(final String filterString)
...
+   *
+   * @return  The current set of parser options.
+   */
+  public static Set<ParserOption> getOptions()
+  {
+    return Collections.unmodifiableSet(Parser.threadLocalOptions.get());
+  }
+
+  /**
+   * Replace the current set of parser options.
```
davidely-ping, line 238: Like with the other methods, I'd mention that this is just for the current thread.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/Parser.java
```diff
@@ -161,6 +177,76 @@ public static Filter parseFilter(final String filterString)
...
+   *
+   * @param options  The updated set of parser options.
+   */
+  public static void setOptions(final Set<ParserOption> options)
+  {
+    Set<ParserOption> updatedOptions = Parser.threadLocalOptions.get();
+    updatedOptions.clear();
+    updatedOptions.addAll(options);
+    Parser.threadLocalOptions.set(updatedOptions);
```
davidely-ping, line 247: This line isn't necessary since updatedOptions is what is already set.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/Parser.java
```diff
@@ -161,6 +177,76 @@ public static Filter parseFilter(final String filterString)
...
+   * @param options  The parser options to be removed.
+   */
+  public static void removeOptions(final ParserOption... options)
+  {
+    Set<ParserOption> updatedOptions = Parser.threadLocalOptions.get();
+    for (ParserOption option : options)
+    {
+      updatedOptions.remove(option);
+    }
+    Parser.threadLocalOptions.set(updatedOptions);
```
davidely-ping, line 224: If we keep this code, then this line can be removed since we're editing the map in place.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/Parser.java
```diff
@@ -161,6 +177,76 @@ public static Filter parseFilter(final String filterString)
...
+   * @param options  The parser options to be added.
+   */
+  public static void addOptions(final ParserOption... options)
+  {
+    Set<ParserOption> updatedOptions = Parser.threadLocalOptions.get();
+    for (ParserOption option : options)
+    {
+      updatedOptions.add(option);
+    }
+    Parser.threadLocalOptions.set(updatedOptions);
```
davidely-ping, line 208: Tthis line can be removed since we're editing the map in place.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/Parser.java
lance-purple-unboundid: So the following?

```
Set<ParserOption> priorOptions = null;
try
{
  priorOptions = Parser.addOptions(ParserOption.ALLOW_SEMICOLONS_IN_ATTRIBUTE_NAMES);
  performWhateverProcessing();
}
finally
{
  if (null != priorOptions)
  {
    Parser.setOptions(priorOptions);
  }
}
```

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/Parser.java
```diff
@@ -161,6 +177,65 @@ public static Filter parseFilter(final String filterString)
...
+   * thread.
+   *
+   * <p>NOTE: SCIM server implementations are not guaranteed to support a given option.</p>
+   * <p>NOTE: These should be reset when no longer needed.</p>
+   *
+   * <pre>
+   *   Set&lt;ParserOption&gt; priorOptions = null;
+   *   try
+   *   {
+   *     priorOptions = Parser.addOptions(ParserOption.ALLOW_SEMICOLONS_IN_ATTRIBUTE_NAMES);
```
davidely-ping, line 191: I'd move this out of the try block so you don't have to check for null in the finally block. Having things like this outside of the try block is a more typical pattern, like for grabbing a lock and releasing it.

lance-purple-unboundid: OK
```
 Set<ParserOption> priorOptions =
     Parser.addOptions(ParserOption.ALLOW_SEMICOLONS_IN_ATTRIBUTE_NAMES);
 try
 {
   performWhateverProcessing();
 }
 finally
 {
   Parser.setOptions(priorOptions);
 }
 ```

**Comment by braveulysses** [comment]
I wonder if we could try a slightly different approach? Using an enum to configure the Parser might limit us if a future need to custom-configure it arises:

1. If you want to allow some character besides semicolon in the future, you have to define another enum value just for that character.
2. ParserOptions are on/off flags. If you want to pass configuration to a ParserOption... I'm not sure that you could?

Maybe you could use something like a ParserConfig class, instead...

```java
public final class ParserConfig
{
  private final Set<Character> allowedCharacters = new HashSet<>();

  public void setAdditionalAllowedCharacters(final Character... allowedCharacters)
  {
    Arrays.stream(allowedCharacters).filter(Objects::nonNull).forEach(this.allowedCharacters::add);
  }

  public void clearAdditionalAllowedCharacters()
  {
    allowedCharacters.clear();
  }

  public Set<Character> getAdditionalAllowedCharacters()
  {
    return allowedCharacters;
  }

  // It should be possible to use a regex Pattern to define the allowed characters instead 
  // of a Set<Character>, too

  // Other methods can be added as needed in the future
}
```

... then assign if to a ThreadLocal in Parser...

```java
private static final ThreadLocal<ParserConfig> threadLocalConfig =
      ThreadLocal.withInitial(ParserConfig::new);
```

... and expose it to the API user with a `public static ParserConfig getParserConfig()` method on the Parser class.  

```java
public static ParserConfig getConfiguration()
{
  return Parser.threadLocalConfig.get();
}
```

Internally, the Parser class would call `Parser.threadLocalConfig.get().getAdditionalAllowedCharacters()` to get the set of additional allowed characters, if any.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/Parser.java
```diff
@@ -322,6 +368,10 @@ private static String readPathToken(final StringReader reader)
       {
         b.append((char)c);
       }
+      else if ((c == ';') && allowSemicolonsInAttributeNames())
```
braveulysses, line 371: Instead of an `allowSemicolonsInAttributeNames()` method, I'd make an `allowedExtendedCharacters()` convenience method like this:

```java
private static Set<Character> allowedExtendedCharacters()
{
  Parser.threadLocalOptions.get().getExtendedAttributeNameCharacters();
}
```

and then change the if condition to something like

```java
allowedExtendedCharacters().contains((char) c)
```

Otherwise, the ParserOptions mechanism will _only_ work if the caller configures it to allow the semicolon character, but it won't work for any other set of characters.

I think you could also get rid of this else-if block and just add the condition to the if block right above this? I.e.,

```java
if (c == '-' || c == '_' || c == '$' || Character.isLetterOrDigit(c) || allowedExtendedCharacters().contains((char) c))
{
...
```

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/Parser.java
```diff
@@ -430,6 +480,10 @@ private static String readFilterToken(final StringReader reader,
...
```
braveulysses, line 483: See my above comment on line 371.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/Parser.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2020 Ping Identity Corporation
+ * Copyright 2015-2021 Ping Identity Corporation
```
braveulysses, line 2: Instead of updating copyright statements piecemeal, we usually update entire repos at the beginning of each year using Neil's copyright tool. Can you please do that in a separate MR?

braveulysses (changes requested):

braveulysses:
One other thing: Can you add a blurb about this change to [CHANGELOG.md](https://github.com/pingidentity/scim2/blob/add-filter-parser-options/CHANGELOG.md)?

braveulysses (approved):

PR #161, opened 2021-07-07
Fix issue where RequestBuilder.toScimException() gives unhelpful erro…

lance-purple-unboundid:
…r details

JIRA issue: DS-44767

### DIFF
--- a/scim2-sdk-client/src/main/java/com/unboundid/scim2/client/requests/RequestBuilder.java
+++ b/scim2-sdk-client/src/main/java/com/unboundid/scim2/client/requests/RequestBuilder.java
@@ -183,18 +183,13 @@ public class RequestBuilder<T extends RequestBuilder>
     }
     catch(ProcessingException ex)
     {
-      String errorMessage = ex.getMessage();
-
-      if (ex.getClass().getSimpleName().equals("MessageBodyProviderNotFoundException"))
-      {
-        // The exception error message contains unwanted details about an internal
-        // Jersey failure to parse the response, rather than the actual SCIM issue.
-        // Replace it with a standard reason phrase for the status code.
-        errorMessage = response.getStatusInfo().getReasonPhrase();
-      }
+      // The exception message likely contains unwanted details about why the
+      // server failed to process the response, instead of the actual SCIM
+      // issue. Replace it with a general reason phrase for the status code.
+      String genericDetails = response.getStatusInfo().getReasonPhrase();
 
       return new ScimServiceException(
-          response.getStatus(), errorMessage, ex);
+          response.getStatus(), genericDetails, ex);
     }
   }
 
--- a/scim2-sdk-client/src/main/java/com/unboundid/scim2/client/requests/RequestBuilder.java
+++ b/scim2-sdk-client/src/main/java/com/unboundid/scim2/client/requests/RequestBuilder.java
@@ -178,12 +178,23 @@ public class RequestBuilder<T extends RequestBuilder>
         ScimException.createException(response.getStatus(), null) :
         ScimException.createException(errorResponse, null);
       response.close();
+
       return exception;
     }
     catch(ProcessingException ex)
     {
+      String errorMessage = ex.getMessage();
+
+      if (ex.getClass().getSimpleName().equals("MessageBodyProviderNotFoundException"))
+      {
+        // The exception error message contains unwanted details about an internal
+        // Jersey failure to parse the response, rather than the actual SCIM issue.
+        // Replace it with a standard reason phrase for the status code.
+        errorMessage = response.getStatusInfo().getReasonPhrase();
+      }
+
       return new ScimServiceException(
-          response.getStatus(), ex.getMessage(), ex);
+          response.getStatus(), errorMessage, ex);
     }
   }
 
--- a/scim2-sdk-server/src/test/java/com/unboundid/scim2/server/EndpointTestCase.java
+++ b/scim2-sdk-server/src/test/java/com/unboundid/scim2/server/EndpointTestCase.java
@@ -1107,9 +1107,13 @@ public class EndpointTestCase extends JerseyTestNg.ContainerPerClassTest
     assertNotNull(user.getMeta().getResourceType());
   }
 
-
   /**
-   * Test bad/unexpected responses from SCIM server.
+   * Ensure that for a failed request, if the returned entity could not be
+   * deserialized as an {@code ErrorResponse}, a {@code ScimServiceException}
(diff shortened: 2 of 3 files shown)


braveulysses (changes requested):
This looks good, and it looks like it will work. I'm mulling over whether a different approach might be better, so I might have additional comments after I get out of some meetings.

In the meantime, I see that you added some tests in the product repository, which is great. We should also add a test here in the SCIM 2 SDK repo. Because of the nature of the issue, it probably makes the most sense to add an integration test in the scim2-sdk-server module. Here's an example:

```patch
Index: scim2-sdk-server/src/test/java/com/unboundid/scim2/server/TestResourceEndpoint.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/scim2-sdk-server/src/test/java/com/unboundid/scim2/server/TestResourceEndpoint.java b/scim2-sdk-server/src/test/java/com/unboundid/scim2/server/TestResourceEndpoint.java
--- a/scim2-sdk-server/src/test/java/com/unboundid/scim2/server/TestResourceEndpoint.java	(revision e59e180e4a7cf5599cf517c3a94062a7962eecf6)
+++ b/scim2-sdk-server/src/test/java/com/unboundid/scim2/server/TestResourceEndpoint.java	(date 1625679639737)
@@ -34,6 +34,7 @@
 import javax.ws.rs.core.Context;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
+import javax.ws.rs.core.Response.Status;
 import javax.ws.rs.core.UriInfo;
 
 import static com.unboundid.scim2.common.utils.ApiConstants.MEDIA_TYPE_SCIM;
@@ -80,6 +81,19 @@
             "}").build();
   }
 
+  /**
+   * TODO: Javadoc
+   */
+  @POST
+  @Path("emptyResponseWithApplicationOctetStreamMediaType")
+  @Produces(MediaType.APPLICATION_OCTET_STREAM)
+  public Response getEmptyResponseWithApplicationOctetStreamMediaType()
+  {
+    return Response.status(Status.UNAUTHORIZED)
+        .type(MediaType.APPLICATION_OCTET_STREAM
+        ).build();
+  }
+
   /**
    * Test SCIM search.
    *
Index: scim2-sdk-server/src/test/java/com/unboundid/scim2/server/EndpointTestCase.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/scim2-sdk-server/src/test/java/com/unboundid/scim2/server/EndpointTestCase.java b/scim2-sdk-server/src/test/java/com/unboundid/scim2/server/EndpointTestCase.java
--- a/scim2-sdk-server/src/test/java/com/unboundid/scim2/server/EndpointTestCase.java	(revision e59e180e4a7cf5599cf517c3a94062a7962eecf6)
+++ b/scim2-sdk-server/src/test/java/com/unboundid/scim2/server/EndpointTestCase.java	(date 1625680056830)
@@ -1132,6 +1132,32 @@
     }
   }
 
+  /**
+   * TODO: Javadoc
+   */
+  @Test
+  public void testEmptyResponseWithApplicationOctetStreamMediaType() throws Exception
+  {
+    final ScimService service = new ScimService(target());
+
+    try
+    {
+      service.createRequest("Users/emptyResponseWithApplicationOctetStreamMediaType",
+          new GenericScimResource())
+          .accept(MediaType.WILDCARD)
+          .invoke();
+      Assert.fail("Expecting a ScimServiceException");
+    }
+    catch(ScimServiceException ex)
+    {
+      ErrorResponse response = ex.getScimError();
+      Assert.assertEquals(response.getStatus(), Integer.valueOf(401));
+      Assert.assertNotNull(ex.getCause());
+      // TODO: Check for an expected error message
+    }
+  }
+  
+  // TODO: Test a few other scenarios, such as text/plain or text/html error responses
 
   /**
    * Test that MethodNotAllowedExceptions are thrown properly.
```

scim2-sdk-client/src/main/java/com/unboundid/scim2/client/requests/RequestBuilder.java
```diff
@@ -178,12 +178,23 @@ static ScimException toScimException(final Response response)
...
         ScimException.createException(errorResponse, null);
       response.close();
+
       return exception;
     }
     catch(ProcessingException ex)
     {
+      String errorMessage = ex.getMessage();
+
+      if (ex.getClass().getSimpleName().equals("MessageBodyProviderNotFoundException"))
```
braveulysses, line 188: MessageBodyProviderNotFoundException is a Jersey exception, so this is unlikely to work if the client is backed by some other JAX-RS client library, such as RestEasy. We want to avoid any runtime dependencies on specific JAX-RS implementations. In the short term, I think we're better off removing this conditional so that this applies to any kind of ProcessingException. 

The problem is that some SDK users might want to know if a deserialization problem or similar problem has occurred. For example, they could be testing a SCIM 2 server implementation, and they actually do want the server to send back a proper SCIM error response. In the long term, we can consider updating the RequestBuilder so that the SDK user can decide how the client will handle cases like this.

It might also make sense for us to start using something like SLF4J in the SDK, so that SDK users have a way to optionally capture exception details that would otherwise get swallowed.

braveulysses (changes requested):

scim2-sdk-server/src/test/java/com/unboundid/scim2/server/EndpointTestCase.java
```diff
@@ -1130,6 +1130,27 @@ public void testBadErrorResult() throws Exception
       Assert.assertEquals(response.getStatus(), Integer.valueOf(409));
       Assert.assertNotNull(ex.getCause());
     }
+
+    // ensure that for a failed request, if the returned entity could not
+    // be deserialized as an ErrorResponse, the SCIM exception contains a
+    // message and a fallback ErrorResponse describing the request failure
+    // NOT the deserialization failure - see DS-44767
```
braveulysses, line 1137: This comment and the fix itself actually apply to both cases exercised by this test method. Why don't you move it up to the top of the method or update the Javadoc?

I'd also update both cases so that they perform the same checks. E.g.:

* ErrorResponse.getStatus returns the expected HTTP status code
* ScimServiceException.getCause does not return a null
* ScimServiceException.getMessage equals ErrorResponse.getDetail
* ErrorResponse.getDetail returns the expected value

I suppose you could even parameterize the test to ensure consistency if you ever want to add other test cases.

braveulysses (approved):

Issue #179, opened 2022-11-30
javax to jakarta migration


MartinHaeusler:
Hello everyone,

since the newly released Spring Boot 3 is now fully based on the new jakarta.* namespaces, it would be great if this package could be upgraded to use jakarta as well.

Thanks!

Fabio1988:
#170 
#173

kqarryzada:
We plan to address this issue within the next couple of months (ideally sooner, but I cannot make any promises about that). Implementing this change will involve extra uplift on our part, since Spring Boot doesn't just require Jakarta EE, it actually requires Jakarta EE 9 (citation: https://spring.io/blog/2021/09/02/a-java-17-and-jakarta-ee-9-baseline-for-spring-framework-6). Some time ago, I did some initial work to migrate our codebase to Jakarta EE 8, but I abandoned this when I realized that it was insufficient.

Thus, this change will require that the latest release of the SCIM SDK drops support for JDK 8. We are in the process of doing this, but we need to work out the logistics since some Ping Identity products still support this JDK.

Thanks for the continued interest in this!

kqarryzada:
Jakarta EE is now supported in the 3.0.0 release, which is available now.

PR #173, opened 2022-09-09
Replace javax package to jakarta package

yokotaso:
_Please be aware that Ping Identity does not accept third-party contributions at this time! Please see our [contribution guidelines](https://github.com/pingidentity/scim2/blob/master/CONTRIBUTING.md)._

What does this implement/fix? Explain your changes.
---------------------------------------------------
It is recommended that javax.xml is replaced by jakarta.xml
https://blogs.oracle.com/javamagazine/post/transition-from-java-ee-to-jakarta-ee

Does this close any currently open issues?
------------------------------------------
None.

PR #170, opened 2022-05-12
Upgrade some dependencies and move jax-rs to jakarta-rs 

otlg:
What does this implement/fix? Explain your changes.
---------------------------------------------------
Move jax-rs to jakarta-rs. It can be merged to master, but I put it here for reference if anyone is interested. 

Does this close any currently open issues?
------------------------------------------
No

PR #180, opened 2022-12-12
Update ALL javax dependencies to jakarta

Fabio1988:
for all others as reference, because the maintainer of this repo are very slow and do not accept PRs :(

#170 (might be duplicated I see right now, sorry)
#173
#179


Fabio1988:
Need to close, as we needed additional fixes... It seems this organization is not willed to keep things up to date

PR #181, opened 2023-02-09
Fix issue where PATCH requests from Azure AD fail due to case-mismatch

kolehey:
JIRA issue: DS-46964

_Please be aware that Ping Identity does not accept third-party contributions at this time! Please see our [contribution guidelines](https://github.com/pingidentity/scim2/blob/master/CONTRIBUTING.md)._

Fixed an issue where PATCH requests from Azure AD
fail due to a case-mismatch in the op type.

Closes DS-46964

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -65,7 +65,7 @@
     <compileSource>1.8</compileSource>
     <main.basedir>${project.basedir}</main.basedir>
     <ignore.test.failures>false</ignore.test.failures>
-    <jackson.version>2.10.2</jackson.version>
+    <jackson.version>2.14.2</jackson.version>
     <jackson-databind.version>2.10.5.1</jackson-databind.version>
     <jaxb.version>2.3.1</jaxb.version>
     <jax-rs.version>2.0.1</jax-rs.version>
--- a/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/messages/PatchOperation.java
+++ b/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/messages/PatchOperation.java
@@ -54,11 +54,11 @@ import java.util.List;
     property = "op")
 @JsonSubTypes({
     @JsonSubTypes.Type(value = PatchOperation.AddOperation.class,
-        name="add"),
+        name="add", names= {"add", "Add", "ADD"}),
     @JsonSubTypes.Type(value = PatchOperation.RemoveOperation.class,
-        name="remove"),
+        name="remove", names= {"remove", "Remove", "REMOVE"}),
     @JsonSubTypes.Type(value = PatchOperation.ReplaceOperation.class,
-        name="replace")})
+        name="replace", names= {"replace", "Replace", "REPLACE"})})
 public abstract class PatchOperation
 {
   static final class AddOperation extends PatchOperation


pom.xml
```diff
@@ -65,7 +65,7 @@
     <compileSource>1.8</compileSource>
     <main.basedir>${project.basedir}</main.basedir>
     <ignore.test.failures>false</ignore.test.failures>
-    <jackson.version>2.10.2</jackson.version>
+    <jackson.version>2.14.1</jackson.version>
```
dirmgr, line 68: It looks like 2.14.2 is the latest version. Is there a reason that you chose to go with 2.14.1 instead?

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/messages/PatchOperation.java
```diff
@@ -54,11 +54,11 @@
...
 @JsonSubTypes({
     @JsonSubTypes.Type(value = PatchOperation.AddOperation.class,
-        name="add"),
+        name="add", names= {"Add", "ADD"}),
     @JsonSubTypes.Type(value = PatchOperation.RemoveOperation.class,
-        name="remove"),
+        name="remove", names= {"Remove", "REMOVE"}),
     @JsonSubTypes.Type(value = PatchOperation.ReplaceOperation.class,
-        name="replace")})
+        name="replace", names= {"Replace", "REPLACE"})})
```
dirmgr, line 61: The [Javadoc](https://javadoc.io/static/com.fasterxml.jackson.core/jackson-annotations/2.14.2/index.html?com/fasterxml/jackson/annotation/JsonSubTypes.html) isn't super clear about how the `name` and `names` annotation elements interact with each other.

The documented description for the `name` element is:

> Logical type name used as the type identifier for the class, if defined; empty String means "not defined". Used unless names is defined as non-empty.

And the documented description for the `names` element is:

> optional) Logical type names used as the type identifier for the class: used if more than one type name should be associated with the same type.

My reading of this is that the two elements are mutually exclusive. The "Used unless names is defined as non-empty" portion of the `name` description suggests that the `name` element is ignored if `names` is defined and non-empty. That's also backed up by the "used if more than one type name should be associated with the same type" portion of the `names` description.

Given that, it sounds like if we're going to use the `names` element, then we want its value to include the complete set of names that should be allowed. In that case, I think the safe thing to do would be to have

`name="add", names={"add", "Add", "ADD"}`

rather than

`name="add", names={"Add", "ADD"}`

And have similar changes for the other operation types.

And if you haven't already, I'd recommend testing your changes against other types of SCIM implementations and not just Azure AD.

pom.xml
kolehey: I checked Nexus to see what the latest version we had there was, and up until I attempted a package with 2.14.2 just now, 2.14.1 was the latest version. 2.14.2 appears to work just fine, so I'll make that change.

dirmgr:
I'm really not very familiar with the intricacies Jackson library (and especially not with its annotations) nor with the SCIM SDK.  I'd recommend also including @vyhhuang on this review.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/messages/PatchOperation.java
kolehey: I'll make sure to check this against other implementations. As for the other operation types, PATCH is the only operation utilizing this JsonSubTypes interface.

dirmgr: Sorry. For "other operation types", I meant the add, remove, and replace operation types within the patch request, although it's also good to know that it's not used for other types of SCIM requests.

vyhhuang (approved):
This looks good to me.

rcereceres (approved):

dougbulkley (approved):

kolehey:
Manual testing was performed by Doug Bulkley, and confirmed that PATCH requests matching Azure AD's case passes.

PR #184, opened 2023-03-03
Update ordering of items within a ListResponse

kqarryzada:
JIRA Issue: DS-46970

### DIFF
--- a/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/messages/ListResponse.java
+++ b/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/messages/ListResponse.java
@@ -19,6 +19,7 @@ package com.unboundid.scim2.common.messages;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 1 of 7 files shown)


scim2-sdk-common/src/main/java/com/unboundid/scim2/common/messages/ListResponse.java
```diff
@@ -51,21 +51,20 @@
   @JsonProperty(value = "totalResults", required = true)
   private final int totalResults;
 
-
```
vyhhuang, line 54: It might be worth using the [JsonPropertyOrder annotation](https://fasterxml.github.io/jackson-annotations/javadoc/2.8/com/fasterxml/jackson/annotation/JsonPropertyOrder.html) instead to order the properties, but since this change works with your unit test I'm fine with it.

vyhhuang (approved):
These changes look good to me; just had a minor comment.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/messages/ListResponse.java
kqarryzada: I went ahead and added that as well.

dougbulkley (approved):

PR #191, opened 2023-06-07
Add varargs helper methods for multi-valued fields

kqarryzada:
The SDK has many methods where an attribute can be set to multiple
values (e.g., setEmails() on UserResource allows a user to have
multiple emails). These methods typically take in an attribute name and
a list of values. There is sometimes some friction with always
requiring a list, however. For example, if a Patch request should have
a single patch operation, then the operation must be wrapped in a
Collections.singletonList() call, which can be clunky.

To remedy this, new methods have been added that allow passing in the
values directly into a method without requiring the caller to wrap the
values in a list first. This is done by adding some overloaded methods
that accept a varargs parameter, and perform the list conversion for
the caller.

These methods always require a single non-null value to avoid breaking
backward compatibility for calls like
'new UserResource().setEmails(null)'.

JiraIssue: DS-47088

### DIFF
--- a/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/BaseScimResource.java
+++ b/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/BaseScimResource.java
@@ -41,6 +41,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import static com.unboundid.scim2.common.utils.StaticUtils.toList;
+
 /**
  * <p>The base SCIM object.  This object contains all of the
  * attributes required of SCIM objects.</p>
@@ -176,7 +178,15 @@ public abstract class BaseScimResource
    */
   public void setSchemaUrns(final Collection<String> schemaUrns)
   {
-    this.schemaUrns = new HashSet<String>(schemaUrns);
+    this.schemaUrns = new HashSet<>(schemaUrns);
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  public void setSchemaUrns(final String schemaUrn, final String... schemaUrns)
+  {
+    setSchemaUrns(toList(schemaUrn, schemaUrns));
   }
 
   /**
--- a/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/BaseScimResource.java
+++ b/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/BaseScimResource.java
@@ -35,10 +35,11 @@ import com.unboundid.scim2.common.utils.SchemaUtils;
 
 import java.util.Collection;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.Iterator;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 
 import static com.unboundid.scim2.common.utils.StaticUtils.toList;
@@ -76,7 +77,7 @@ public abstract class BaseScimResource
   private Meta meta;
 
   @JsonProperty("schemas")
-  private Set<String> schemaUrns = new HashSet<String>();
+  private Set<String> schemaUrns = new LinkedHashSet<>();
 
   private final ObjectNode extensionObjectNode =
       JsonUtils.getJsonNodeFactory().objectNode();
@@ -165,11 +166,6 @@ public abstract class BaseScimResource
    */
   public Set<String> getSchemaUrns()
   {
-    if(schemaUrns == null)
-    {
-      schemaUrns = new HashSet<String>();
-    }
-
     return schemaUrns;
   }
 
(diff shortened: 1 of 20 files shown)


scim2-sdk-common/src/main/java/com/unboundid/scim2/common/GenericScimResource.java
```diff
@@ -469,6 +479,42 @@ public String toString()
...
+  /**
+   * Indicates whether the provided object is equal to this generic SCIM
+   * resource.
+   *
+   * @param o   The object to compare.
+   * @return    {@code true} if the provided object is equal to this generic
+   *            SCIM resource, or {@code false} if not.
+   */
+  @Override
+  public boolean equals(final Object o)
```
kqarryzada, line 491: I added this since I needed it in the unit tests. It felt wrong to use `assertEquals(first.getObjectNode(), second.getObjectNode())`. I also added a test for this method too.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/GenericScimResource.java
```diff
@@ -585,6 +631,24 @@ public GenericScimResource addStringValues(final String path,
...
+   *
+   * @param path    The path to the attribute that should be updated.
+   * @param value1  The first value. This must not be {@code null}.
+   * @param values  An optional field for additional values. Any {@code null}
+   *                values will be ignored.
+   * @return        This object.
+   *
+   * @throws ScimException  If the path is invalid.
+   */
+  public GenericScimResource addStringValues(final String path,
```
kqarryzada, line 645: One thing I noticed is that this file (and `PatchOperation`) have methods for `addBooleanValues()`. Plural. This seems wrong to me. I can't think of a situation where it makes sense to have a multi-valued boolean attribute. At most, it'd ever be `[ true, false ]` (if there are no repeating values), and that array itself doesn't make sense to me.

I can consider adding a `addBooleanValue(String path, boolean value)` method and deprecating the old one, but I thought I'd point this out first.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/ScimResource.java
```diff
@@ -76,11 +76,22 @@
   /**
    * Sets the schema urns for this object.  This set should contain
    * all schema urns including the one for this object and all
-   * extensions.
-   * @param schemaUrns a set containing the schema urns for this object.
+   * extensions.  The value must not be {@code null}.
```
kqarryzada, line 79: I added a unit test verifying this behavior, so it seemed worthwhile to add it in the documentation.

scim2-sdk-common/src/test/java/com/unboundid/scim2/common/UserResourceVarargsTestCase.java
```diff
@@ -0,0 +1,563 @@
+/*
```
kqarryzada, line 1: This test is very long, and just repeats for every MVA on a SCIM user. It's repetitive and verbose, but I wanted to be sure that a change to just one of these methods didn't break any of the contracts that we're expecting.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/GenericScimResource.java
braveulysses: Yes, that's odd. It does look like `replaceValue(String path, Boolean value)` handles the case of adding or updating a single-valued boolean property, though.

braveulysses (approved):
Looks good! I did have a comment about using `LinkedHashSet` to track schema URNs in `BaseScimResource` — I think that would be worth pursuing.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/GenericScimResource.java
kqarryzada: Thanks, I think I'll move forward with marking this method as deprecated, then. Do you think it would be worth adding a `addBooleanValue` method that just calls `replaceValue`? It may not be obvious to clients that this is the appropriate method when they are looking to add a value.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/BaseScimResource.java
```diff
@@ -176,7 +178,15 @@ public Set<String> getSchemaUrns()
    */
   public void setSchemaUrns(final Collection<String> schemaUrns)
   {
-    this.schemaUrns = new HashSet<String>(schemaUrns);
+    this.schemaUrns = new HashSet<>(schemaUrns);
```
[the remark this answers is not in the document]
kqarryzada: I think using a `LinkedHashSet` is a great idea because it would give clients a reliable way to list certain schemas before others. For example, if a resource has an extension schema, it's understandable to want that field to always be printed second.

I agree that it's more ideal to perform that check in the constructor. I'll make this change.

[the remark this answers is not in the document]
kqarryzada: This ended up being a bit simpler since the existing code already would throw an NPE if `null` was provided, so `schemaUrns` always has a non-null value. It was simplest to make an explicit `Objects.requireNonNull()` call in the constructor.

dougbulkley (approved):

braveulysses (approved):

PR #194, opened 2023-06-16
Allow empty array values for PatchOperations

### LINKED ISSUE #103: ReplaceOperation with an empty-array value is rejected unnecessarily
**Problem behavior**
The library rejects payloads that are valid according to [RFC 7644, Sec. 3.5.2.3. Replace Operation](https://tools.ietf.org/html/rfc7644#section-3.5.2.3). Specifically, it rejects `Operations` whose `value` is the empty array `[]`:

```java
      if(value == null || value.isNull() ||
           ((value.isArray() || value.isObject()) && value.size() == 0))
       {
         throw BadRequestException.invalidSyntax(
             "value field must not be null or an empty container");
       }
```
[PatchOperation.java#L305](https://github.com/pingidentity/scim2/blob/92eb02db4140ca2a81ed19fdbc0ccdb2fcce56a1/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/messages/PatchOperation.java#L305)

**Why this is important**
Okta sends the following payload which fails to be deserialized:
```javascript
{
    "schemas": [
        "urn:ietf:params:scim:api:messages:2.0:PatchOp"
    ],
    "Operations": [
        {
            "op": "replace",
            "path": "members",
            "value": []
        }
    ]
}
```

**Historical context**
The code doing this validation was merged in #26.

**Suggested resolution**
Change the check in `PatchOperation.java` to:
```java
      if(value == null || value.isNull() ||
           (value.isObject() && value.size() == 0))
       {
         throw BadRequestException.invalidSyntax(
             "value field must not be null or an empty object");
       }
```

kqarryzada:
The SDK now allows the construction of PATCH Operations with an empty array value. This is useful when a client wants to delete all existing values of a multi-valued attribute (i.e., with 'replace'), but is available for add operations as well.

JiraIssue: DS-47463
Resolves #103.


CHANGELOG.md
```diff
@@ -4,7 +4,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 
 ## v2.3.9 - TBD
```
kqarryzada, line 6: I think this change warrants bumping the minor version, but I have an open PR that will already do this.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/messages/PatchOperation.java
```diff
@@ -1371,4 +1349,35 @@ public static PatchOperation create(final PatchOpType opType,
...
+   */
+  private static void validateOperationValue(final Path path,
+      final JsonNode value, final PatchOpType type)
+          throws ScimException
+  {
+    if (value == null || value.isNull() ||
+            (value.isObject() && value.size() == 0))
+    {
+      throw BadRequestException.invalidSyntax(
+          "value field must not be null or an empty container");
```
pingcognito, line 1373: Would you update the exception message here to be more specific?

kqarryzada: That makes sense. I've updated it to the following:

> The patch operation value must not be null or an empty object

pingcognito (approved):

Issue #135, opened 2019-10-24
PatchOperation$AddOperation instance creation fails on path values with selection filters


KoichiSenada:
**Describe the bug**
PatchOperation.AddOperation instance creation fails on path values with selection filters.
https://github.com/pingidentity/scim2/commit/4b3d63c7c22ec044f5da4bb14848ac1366a3fa13#r35642158
```
com.unboundid.scim2.common.exceptions.BadRequestException: path field for add operations must not include any value selection filters
    at com.unboundid.scim2.common.exceptions.BadRequestException.invalidPath(BadRequestException.java:197)
    at com.unboundid.scim2.common.messages.PatchOperation$AddOperation.<init>(PatchOperation.java:101)
```

**To Reproduce**
Send a SCIM2 protocol https://tools.ietf.org/html/rfc7644#section-3.5.2 compliant PATCH request.
```
{
  "schemas" : [ "urn:ietf:params:scim:api:messages:2.0:PatchOp" ],
  "Operations" : [ {
    "op" : "add",
    "path" : "addresses[type eq \"work\"].country",
    "value" : "Australia"
  }, {
    "op" : "add",
    "path" : "addresses[type eq \"work\"].locality",
    "value" : "Australia"
  }, {
    "op" : "add",
    "path" : "addresses[type eq \"work\"].formatted",
    "value" : "Australia"
  } ]
}
```

**Expected behavior**
It is expected that the PatchOperation$AddOperation instance would be created.
https://tools.ietf.org/html/rfc7644#section-3.5.2
```
   Valid examples of "path" are as follows:

       "path":"members"

       "path":"name.familyName"

       "path":"addresses[type eq \"work\"]"

       "path":"members[value eq
              \"2819c223-7f76-453a-919d-413861904646\"]"

       "path":"members[value eq
              \"2819c223-7f76-453a-919d-413861904646\"].displayName"
```

Stummi:
We seem to encounter the same issue when trying to integrate AD. 

Ironically, its just the `AddOperation` Constructor which plainly refuses any path with a value selection filter. Ther underlaying tools can work with the filter without issues:

```kotlin
val body = """
    {
        "addresses": [
        {
            "type": "work",
            "locality": "old"
        }
        ]
    }
""".trimIndent()

val om = ObjectMapper()
val node: ObjectNode = om.readValue(body, ObjectNode::class.java)
val newAddr: JsonNode = om.readValue("\"new\"", JsonNode::class.java)
[...]

PR #195, opened 2023-06-20
Added capability to evaluate path queries in Patch Add operations.

MartinHaeusler:

The SCIM2 framework at the moment does not support path queries in PATCH operations if they occur within an "add" op. For example, the following PATCH operation would be invalid:

```json
{
    "schemas": [
        "urn:ietf:params:scim:api:messages:2.0:PatchOp"
    ],
    "Operations": [
        {
            "op": "Add",
            "path": "addresses[type eq \"work\"].country",
            "value": "Austria"
        }
    ]
}
```

... because of the `[type eq \"work\"]` filter.

Yes: https://github.com/pingidentity/scim2/issues/135


kqarryzada:
This functionality is now available in the 3.0.0 release, so I'm closing this PR.

PR #197, opened 2023-06-30
Upgrade TestNG to 7.5.1

kqarryzada:
Updated the test framework to the latest (JDK 8-compatible) supported
release in order to resolve a CVE warning.

The new TestNG version introduced some inconsistent behavior in the way
that 'assertEquals' behaves in the new release. Now, it appears that
iterables and collections must have their elements in the same order to
be considered equal. TestNG surprisingly does not call an object's
equals() method to determine this. This led to failures in some tests
that checked equivalency of JsonNodes, which should allow elements to
be out of order.

To remedy this, the project has added the AssertJ dependency. This
library helps with writing flexible assertion statements that will be
useful in the future, and correctly compares JsonNodes in an expected
and reliable fashion.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -231,7 +231,7 @@
             <phase>package</phase>
             <configuration>
               <configLocation>config/scim2-parent-checkstyle.xml</configLocation>
-              <sourceDirectory>${project.build.sourceDirectory}</sourceDirectory>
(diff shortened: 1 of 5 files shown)


vyhhuang (approved):

PR #199, opened 2023-07-03
Update Member attribute definitions

### LINKED ISSUE #189: Sub-attributes of group resource's member attribute should be immutable
**Describe the bug**
[According to the RFC](https://www.rfc-editor.org/rfc/rfc7643#section-4.2), the sub-attributes of the `group` resource's `member` attribute should be `immutable` instead of `readOnly`.

**To Reproduce**
Call the `/Schemas` endpoint. The `mutability` is `readOnly` for all sub-attribute of the `member` attribute instead of `immutable`, see the `$ref` attribute, for example:

```JSON
{
  "name": "$ref",
  "type": "reference",
  "multiValued": false,
  "description": "The URI of the corresponding Group resource to which the user belongs",
  "required": false,
  "caseExact": true,
  "mutability": "readOnly",
  "returned": "default",
  "uniqueness": "none",
  "referenceTypes": [
      "Group",
      "User"
  ]
}
```

**Expected behavior**
The values of `mutability` of all the sub-attributes should be `immutable`.

**Additional context**
 - SCIM 2 SDK version: [2.3.8 and master branch]

kqarryzada:
This is broken into two parts. One resolved a typo, and the other included a missing sub-attribute definition.

This change may have to wait for after the 2.4.0 release.

JiraIssue: DS-47600
Resolves #189

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ Updating a boolean value should always be done with a `replace` operation type r
 Fixed an issue where `AttributeDefinition.toString()` would not print the mutability of an
 attribute.
 
+Added a `type` field to the `Member` class as defined by RFC 7643 section 8.7.1.
+
 ## v2.3.8 - 2023-05-17
 Updated the deserialized form of ListResponse objects so that the `itemsPerPage` and `startIndex`
 fields are listed at the top with `totalResults`. This matches the form of ListResponses shown in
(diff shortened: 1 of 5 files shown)


vyhhuang (approved):
These changes look good to me. The values chosen for the new test cases are quite curious.

dougbulkley (approved):

PR #202, opened 2023-07-21
Resolve error with copying SCIM SDK ObjectMapper

kqarryzada:
This commit resolves an error with copying an ObjectMapper that was created with 'JsonUtils.createObjectMapper()'. This presented as an IllegalStateException from Jackson's 'ObjectMapper._checkInvalidCopy()' method.

Part of this commit includes adding a new constructor for ScimJsonFactory since there is a dedicated constructor used for copies in the superclass (JsonFactory#JsonFactory(JsonFactory, ObjectCodec)). ScimJsonFactory will explicitly call this "copy constructor" when making a copy, even though the behavior is (currently) identical to using the default constructor.

JiraIssue: DS-47635


vyhhuang (approved):
Looks good to me!

dougbulkley (approved):

PR #203, opened 2023-08-09
Update empty javadocs

kqarryzada:
Resolved an issue where methods such as toString(), hashCode(), and equals() did not have any descriptions in the public Javadocs. This was due to our use of "inheritDoc", which does not inherit the text from the parent java.lang.Object class unless a class explicitly extends Object. To correct this behavior, this commit adds dedicated descriptions for these fields.


vyhhuang (approved):
This looks good to me.

PR #204, opened 2023-08-21
Update Maven plugins

kqarryzada:
Updated the maven-checkstyle-plugin to resolve a potential issue with
the build pipeline. The version of Checkstyle used by the plugin has
jumped from 6.19 to 9.3, so some rules in the checkstyle.xml were no
longer valid and required updates. The necessary project code changes
to accommodate the new ruleset will be reflected in another commit.

The Maven Assembly plugin also required an update to resolve a pipeline
build failure as a result of the Checkstyle update. To prevent similar
issues, other Maven plugins have been updated to the most recent
releases as well.

### DIFF
--- a/config/scim2-parent-unit-checkstyle.xml
+++ b/config/scim2-parent-unit-checkstyle.xml
@@ -35,19 +35,34 @@
 
 
   <module name="TreeWalker">
-    <property name="cacheFile" value="${cacheFile}"/>
     <!-- Ensure that all methods have Javadoc documentation. -->
     <module name="JavadocMethod">
-      <property name="scope"                       value="package" />
-      <property name="allowUndeclaredRTE"          value="false"   />
+      <property name="accessModifiers"  value="package, protected, public" />
       <property name="allowMissingParamTags"       value="false"   />
-      <property name="allowMissingThrowsTags"      value="false"   />
       <property name="allowMissingReturnTag"       value="false"   />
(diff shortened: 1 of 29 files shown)


config/scim2-parent-unit-checkstyle.xml
```diff
@@ -162,5 +167,18 @@
     <!-- Ensure that all long constants are followed by a capital L. -->
     <module name="UpperEll" />
   </module>
-</module>
 
+  <!-- Ensure that there are no lines longer than 120 characters in the test
+       code. -->
+  <module name="LineLength">
+    <property name="fileExtensions" value="java"/>
+    <property name="max" value="120"/>
```
vyhhuang, line 175: In `config/scim2-parent-checkstyle.xml`, this max value is set to 100. Is it worth changing either this or the other value so they match each other?

vyhhuang (approved):
I had one comment, but feel free to take it or leave it, as it's pretty minor. Aside from that, looks good to me.

config/scim2-parent-unit-checkstyle.xml
kqarryzada: I left this higher because the line-length styling is much less important in the test code in my opinion. It seems okay to give more leniency than 100 characters in the test package. But let me know if you disagree.

vyhhuang: Sure, that sounds fine to me.

PR #205, opened 2023-08-24
Overhaul filter documentation.

kqarryzada:
This commit enhances the Javadocs for the Filter class and its
subclasses with additional background and more detailed descriptions.
The goal is to give enough detail so that SDK clients are not forced to
refer to the RFC for basic SCIM concepts.

JiraIssue: DS-47822

### DIFF
--- a/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/filters/ComparisonFilter.java
+++ b/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/filters/ComparisonFilter.java
@@ -23,10 +23,12 @@ import com.unboundid.scim2.common.utils.JsonUtils;
 
 /**
  * This superclass represents filter types that are used to compare attribute
- * values of SCIM resources. For example, an {@link EqualFilter} is a comparison
- * filter. A filter such as {@code preferredLanguage eq "en-US"} is used to
- * perform a comparison between the filter value, {@code "en-US"}, and the value
- * of a SCIM resource's {@code preferredLanguage} attribute.
+ * values of SCIM resources.
+ * <br><br>
+ * For example, an {@link EqualFilter} is a comparison filter. A filter such as
+ * {@code preferredLanguage eq "en-US"} is used to perform a comparison between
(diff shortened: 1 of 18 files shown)


scim2-sdk-common/src/main/java/com/unboundid/scim2/common/filters/ComparisonFilter.java
```diff
@@ -22,7 +22,14 @@
 import com.unboundid.scim2.common.utils.JsonUtils;
 
 /**
- * Attribute comparison filter.
+ * This superclass represents filter types that are used to compare attribute
+ * values of SCIM resources. For example, an {@link EqualFilter} is a comparison
```
braveulysses, line 26: Nit: Put the example in a separate paragraph. 

My thinking here is that `ComplexValueFilter` establishes a good, readable pattern that would also work well here:

1. Summary/definition
2. Example
3. Supplementary information

This comment — if you choose to accept it — also applies to other classes, such as `ContainsFilter` and `EndsWithFilter`.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/filters/OrFilter.java
```diff
@@ -22,7 +22,32 @@
 import java.util.List;
 
 /**
- * Logical OR combining filter.
+ * This class represents a SCIM {@code or} filter. An OR filter allows
+ * a client to specify two different filter criteria where at least one
+ * of the criterion must match. For instance, consider the following filter.
```
braveulysses, line 27: Should be either "at least one criterion must match" or "at least one of the criteria must match".

braveulysses (changes requested):
Great work, @kqarryzada. Some of my requests for changes are a matter of editorial style, and whether they should be addressed is up to you. 

However, I did see some _existing_ Javadoc problems that should be corrected; I noted those with a comment near the top of `Filter`.

vyhhuang (approved):
This looks good to me.

braveulysses (approved):
Sorry about the late approval, @kqarryzada! LGTM.

PR #207, opened 2023-09-26
Support value selection filters in patch add ops 

### LINKED ISSUE #135: PatchOperation$AddOperation instance creation fails on path values with selection filters

kqarryzada:
Some SCIM service providers embed value selection filters in patch add
operations. The contents of the filter are intended to specify
additional data within the request, which is different from the
behavior of remove and replace operations. Remove and replace operations
use the filter to evaluate the values should be modified/removed.

RFC 7644 explicitly supports the use of value selection filters
the 'path' field of remove and replace operations. It does not say
anything about the use of value selection filters for add operations,
but it does not forbid them, either. Since this form is popular enough,
the SDK no longer rejects operations of these types, and also provides
support for applying these modifications as expected in a patch
request.

Reviewer: vyhhuang
Reviewer: dougbulkley

JiraIssue: DS-47623
Resolves #135

### DIFF
--- a/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/messages/PatchOperation.java
+++ b/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/messages/PatchOperation.java
@@ -32,11 +32,16 @@ import com.fasterxml.jackson.databind.node.IntNode;
 import com.fasterxml.jackson.databind.node.LongNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.fasterxml.jackson.databind.node.TextNode;
+import com.fasterxml.jackson.databind.node.ValueNode;
 import com.unboundid.scim2.common.GenericScimResource;
 import com.unboundid.scim2.common.Path;
+import com.unboundid.scim2.common.annotations.NotNull;
+import com.unboundid.scim2.common.annotations.Nullable;
 import com.unboundid.scim2.common.exceptions.BadRequestException;
 import com.unboundid.scim2.common.exceptions.ScimException;
+import com.unboundid.scim2.common.filters.EqualFilter;
 import com.unboundid.scim2.common.filters.Filter;
+import com.unboundid.scim2.common.filters.FilterType;
 import com.unboundid.scim2.common.utils.JsonUtils;
 import com.unboundid.scim2.common.utils.SchemaUtils;
 
@@ -124,18 +129,6 @@ public abstract class PatchOperation
     {
       super(path);
       validateOperationValue(path, value, getOpType());
-      if (path != null)
-      {
-        for (Path.Element element : path)
-        {
-          if (element.getValueFilter() != null)
-          {
-            throw BadRequestException.invalidPath(
-                "path field for add operations must not include any value " +
-                    "selection filters");
-          }
-        }
-      }
       this.value = value;
     }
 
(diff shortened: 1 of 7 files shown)


scim2-sdk-common/src/main/java/com/unboundid/scim2/common/messages/PatchOperation.java
```diff
@@ -153,11 +146,201 @@ public <T> List<T> getValues(final Class<T> cls)
...
-          getPath(), node, value);
+      Path path = (getPath() == null) ? Path.root() : getPath();
+      if (hasValueFilter(path))
+      {
+        validateAddOpWithFilter(path, value);
+        applyAddWithValueFilter(path, node, value);
+      }
+      else
+      {
+        JsonUtils.addValue(path, node, value);
```
kqarryzada, line 157: I tried really hard to make these changes follow the Visitor pattern in JsonUtils by adding a `CreatingNodeVisitor` type. However, I wasn't able to get this to work because the "traversal" nature of `traverseValues()` didn't play well with this. When someone typically wants to add a new value to an array, the value in the patch operation is set to an array of a single element. But patch requests with value selection filters don't set the value to an array. Hence, I had to resort to an if-statement to perform alternate logic for cases where a value selection filter is present.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/messages/PatchOperation.java
```diff
@@ -153,11 +146,201 @@ public <T> List<T> getValues(final Class<T> cls)
...
+      if (filterType != FilterType.EQUAL)
+      {
+        throw BadRequestException.invalidPath(String.format(
+            "The add operation contained a value selection filter of type '%s',"
+                + " which is not an equality filter",
+            filterType));
+      }
+      if (!it.hasNext())
+      {
+        // A Path with a filter must have a second element to specify the
```
vyhhuang, line 228: I think it's worth including this criteria in the method comment as well.

scim2-sdk-common/src/test/java/com/unboundid/scim2/common/AddOperationValueFilterTestCase.java
```diff
@@ -0,0 +1,288 @@
...
+  @Test
+  public void testDeserializedObject() throws Exception
+  {
+    final String rawExpectedResult = "{"
+        + "  \"schemas\" : [ \"urn:ietf:params:scim:schemas:core:2.0:User\" ],"
+        + "  \"emails\" : [ {"
+        + "    \"value\" : \"sissel@example.com\","
+        + "    \"type\" : \"work\""
+        + "  }, {"
+        + "    \"value\" : \"sissel2@example.com\","
```
vyhhuang, line 247: I think "yomiel@example.com" could be used here.

vyhhuang (approved):
These changes look good to me. Feel free to disregard my second comment if you want, haha.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/messages/PatchOperation.java
kqarryzada: Good idea. I've added this.

scim2-sdk-common/src/test/java/com/unboundid/scim2/common/AddOperationValueFilterTestCase.java
kqarryzada: Excellent suggestion; done!

dougbulkley (approved):
Nice tests, thanks.

PR #211, opened 2023-10-11
Fix scim2-sdk-server tests being skipped

kqarryzada:
Added some test dependencies that were removed in a previous commit, as they were incorrectly determined to be unused. These changes alone were insufficient because the newest version of the Maven Surefire Plugin was not correctly invoking the TestNG framework. To work around this issue, the plugin has been explicitly configured to use TestNG.


vyhhuang (approved):
These changes seem good to me.

PR #212, opened 2023-10-18
Add GitHub Actions build pipeline. 

kqarryzada:
This commit adds a basic build pipeline for the default branch as well
as on pull requests. Existing references to TravisCI have been removed,
as this tool is no longer being used.

Reviewer: henryrecker-pingidentity
Reviewer: dougbulkley

JiraIssue: DS-48068


.github/workflows/build-and-test.yaml
```diff
@@ -0,0 +1,55 @@
...
+    - name: Install minimum-supported JDK for Checkstyle
+      uses: actions/setup-java@v3
+      with:
+        java-version: 11
+        distribution: 'temurin'
+        cache: maven
+
+    - name: Checkstyle
+      run: |
+        ./mvnw checkstyle:check -Dcheckstyle.config.location=config/scim2-parent-checkstyle.xml
```
kqarryzada, line 27: I added a [maven wrapper](https://maven.apache.org/wrapper/) so that we have firm control over the version of Maven that is used by these images.

.github/workflows/build-and-test.yaml
```diff
@@ -0,0 +1,55 @@
...
+    strategy:
+      matrix:
+        # Compile and build with supported LTS releases of Java.
+        java-version: [ 11, 17, 21 ]
+
+    steps:
+    - run: |
+        echo "This job was triggered by the '${{ github.event_name }}' event."
+    - run: |
+        echo "The branch name is '${{ github.ref }}'."
```
henryrecker-pingidentity, line 44: This information is probably available on the running action, so you may not need to echo this.

.github/workflows/build-and-test.yaml
henryrecker-pingidentity: Seems like you have to add quite a lot to the repo (the jar file, etc) to get this wrapper. What's the cost of not doing this and just using some built in maven action?

.github/workflows/build-and-test.yaml
kqarryzada: Good call, it is there. I've removed this.

.github/workflows/build-and-test.yaml
kqarryzada: My concern about this is that I don't think there's a way to control the version of Maven that the pipeline pulls in. I could be wrong, but it would be suboptimal if the pipeline starts failing due to an incompatibility with a future version of the Maven client. For what it's worth, using Maven wrappers and committing them to source control is not uncommon.

henryrecker-pingidentity (approved):

PR #217, opened 2024-03-13
Support replace ops with an empty array value

### LINKED ISSUE #209: `ReplaceOperation` is not applied correctly when value is an empty array
**Describe the bug**
Applying a `ReplaceOperation` with an empty array value does not have any effect.

**To Reproduce**
```java
package scim2libbug;

import com.fasterxml.jackson.databind.node.JsonNodeFactory;
import com.unboundid.scim2.common.exceptions.ScimException;
import com.unboundid.scim2.common.messages.PatchOperation;
import com.unboundid.scim2.common.types.Email;
import com.unboundid.scim2.common.types.UserResource;

public class Main {
    public static void main(String[] args) throws ScimException {
        var scimUser = new UserResource();
        var workEmail = new Email();
        workEmail.setPrimary(true);
        workEmail.setType("work");
        workEmail.setValue("test@email.com");
        scimUser.setEmails(workEmail);
        var genericUser = scimUser.asGenericScimResource();
        var patchOp = PatchOperation.replace("emails", JsonNodeFactory.instance.arrayNode());

        System.out.println("BEFORE:");
        System.out.println(genericUser);
        patchOp.apply(genericUser.getObjectNode());
        System.out.println("AFTER:");
        System.out.println(genericUser);
    }
}

/* Output:
BEFORE:
{
  "schemas" : [ "urn:ietf:params:scim:schemas:core:2.0:User" ],
  "emails" : [ {
    "value" : "test@email.com",
    "type" : "work",
    "primary" : true
  } ]
}
AFTER:
{
  "schemas" : [ "urn:ietf:params:scim:schemas:core:2.0:User" ],
  "emails" : [ {
    "value" : "test@email.com",
    "type" : "work",
    "primary" : true
  } ]
}
*/
```

**Expected behavior**
In the above example, the emails should be set to an empty list. I believe this was the behaviour that was intended to work in https://github.com/pingidentity/scim2/issues/103

**Additional context**
Looks like there's a condition in `JsonUtils` that's preventing the operation from applying:
https://github.com/pingidentity/scim2/blob/7dbeb1cacbebe48ccc5b221d3260153b3ae41160/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/JsonUtils.java#L319-L326

We're currently working around this for now by clearing array fields before applying our patch operations.

kqarryzada:
This commit resolves an issue with replace operations that contained an empty array as the value. If a filter is not included in the path, the SDK now treats this as a request to delete all values of the targeted multi-valued attribute. If a filter is included, then all attribute values that match the filter will be removed.

Reviewer: vyhhuang
Reviewer: dougbulkley

JiraIssue: DS-48754
Resolves #209

### DIFF
--- a/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/JsonUtils.java
+++ b/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/JsonUtils.java
@@ -28,6 +28,8 @@ import com.fasterxml.jackson.databind.node.NullNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.fasterxml.jackson.databind.type.CollectionType;
 import com.unboundid.scim2.common.Path;
+import com.unboundid.scim2.common.annotations.NotNull;
+import com.unboundid.scim2.common.annotations.Nullable;
 import com.unboundid.scim2.common.exceptions.BadRequestException;
 import com.unboundid.scim2.common.exceptions.ScimException;
 import com.unboundid.scim2.common.filters.Filter;
@@ -35,6 +37,7 @@ import com.unboundid.scim2.common.messages.PatchOperation;
 import com.unboundid.scim2.common.types.AttributeDefinition;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Date;
 import java.util.Iterator;
 import java.util.LinkedList;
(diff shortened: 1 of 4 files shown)


scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/JsonUtils.java
```diff
@@ -254,55 +257,131 @@ else if(node.isArray())
...
+          indexesToDelete.add(i);
+        }
+        else
+        {
+          array.set(i, value);
         }
       }
-      updateNode(parent, field, value);
+
+      for (int j = indexesToDelete.size() - 1; j >= 0; j--)
```
vyhhuang, line 369: I think this looks good and doesn't necessarily need to change, but it seems possible to avoid the second loop for 'indexesToDelete' if we just go backwards through the array in the first loop. I think keeping the 'update' and 'delete' steps separated out might just be a cleaner implementation, though.

vyhhuang (approved):
I had one minor comment but it really doesn't require any action. These changes look good to me.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/JsonUtils.java
kqarryzada: That's a good idea, I didn't think of that. I initially had the main loop delete the index and then decrement `i` to account for the shifting, but I agree that it seems cleaner to just separate the two steps.

dougbulkley (approved):

PR #220, opened 2024-04-19
Add Nullability Annotations.

kqarryzada:
This PR updates the SCIM SDK to utilize the `@NotNull` and `@Nullable` annotations on member variables, input parameters, and method return types to match the form of the LDAP SDK. 

This PR is split up into multiple commits due to the volume of changes. I would recommend reviewing this one commit at a time. The purpose of the first three commits are:
1. The initial copy of the LDAP SDK's unit test to ensure that these annotations are maintained for all future commits
2. All updates required to make the test work with the SCIM SDK's `common` module
3. Alternate versions of the tests for the other modules, e.g., `scim2-sdk-client`

I tried very hard to avoid copying and pasting the test into all of the separate modules. I initially came up with a solution that built a test-jar of the scim2-sdk-common tests so that the other modules could extend the NullabilityAnnotationTest as a parent class. But this added complexity, because all of the modules now depended on the common module's tests being run first. This is more akin to an integration test, but adding the Maven plugin to support this as an integration test seemed like too much effort when this test isn't a true integration test anyway. In the end, I decided to just copy the test, since it's unlikely that it will be changed much at all, so upkeep is not really a major concern.


braveulysses (approved):
LGTM, @kqarryzada. Since this is so large, I didn't verify that each use of these annotations was correct, but instead sampled several key classes, such as `BaseScimResource`. Even then, I may have missed something, but I think that's OK. If there are any mistakes, they won't block developers who use the SCIM 2 SDK, since there is no compile-time enforcement of these annotations.

Annotations like these are arguably most effective when they're supported by the developer's IDE, but I doubt that most developers know how to configure this IDE support. So it might be helpful to add a brief section to the README with pointers. 

* IntelliJ IDEA: https://www.jetbrains.com/help/idea/annotating-source-code.html
* Eclipse: https://help.eclipse.org/latest/index.jsp?topic=%2Forg.eclipse.jdt.doc.user%2Ftasks%2Ftask-using_null_annotations.htm

kqarryzada:
Thanks for the suggestion, @braveulysses. I set this up locally and it helped identify some mistakes. I'll add a README update explaining this soon.

markbernhardt-ping (approved):

README.md
```diff
@@ -114,6 +114,19 @@ ListResponse<UserResource> searchResponse =
 
 For detailed information about using the SCIM 2 SDK, including more examples, please see the [wiki](https://github.com/pingidentity/scim2/wiki).
 
+# Nullability Annotations
```
braveulysses, line 117: Thanks for adding this — looks great!

vyhhuang (approved):
The changes look good to me. I was going to comment on the tests being duplicated, but I saw your message and your reasoning makes sense.

PR #223, opened 2024-05-17
Eliminate warnings.

kqarryzada:
Updated some SDK code to avoid warning messages from code scanning tools. This includes a change to the StringReader#skip method, which now explicitly checks for negative values.

JiraIssue: DS-48860

### DIFF
--- a/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/Parser.java
+++ b/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/Parser.java
@@ -50,6 +50,7 @@ public class Parser
   {
     @NotNull
     private final String string;
+
     private int pos;
     private int mark;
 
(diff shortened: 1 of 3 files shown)


scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/Parser.java
```diff
@@ -122,12 +123,23 @@ public void reset()
...
      */
     @Override
-    public long skip(final long n)
+    public long skip(final long n) throws IllegalArgumentException
     {
+      if (n < 0L)
+      {
+        throw new IllegalArgumentException(
+            "Attempted to skip a negative number of characters.");
+      }
```
kqarryzada, line 136: I needed to ensure that the input value was not a very large negative number. The parent class throws an IllegalArgumentException for any negative value, so it made sense to include that here too.

kqarryzada:
I've left details about testing this change on Jira.

dougbulkley (approved):

vyhhuang (approved):

PR #228, opened 2024-07-19
Support customization of the MapperFactory 

### LINKED ISSUE #147: Json customization via MapperFactory doesn't work
**Describe the bug**
It is impossible to customize JSON serialization/deserialization using https://github.com/pingidentity/scim2/blob/master/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/MapperFactory.java and https://github.com/pingidentity/scim2/blob/master/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/JsonUtils.java#L1049

**To Reproduce**
Calling https://github.com/pingidentity/scim2/blob/master/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/JsonUtils.java#L1049 will not have any useful effect since https://github.com/pingidentity/scim2/blob/master/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/MapperFactory.java#L139 method is static and linked at compile time. 
Thus https://github.com/pingidentity/scim2/blob/master/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/JsonUtils.java#L1052 will always call `createObjectMapper` of `MapperFactory` and passed `customMapperFactory` will be effectively ignored.

**Expected behavior**
` SDK_OBJECT_MAPPER` should be assigned new value from `customMapperFactory`

**Additional context**
Right now it is incompatible with Okta's version of SCIM messages. Namely, fields `roles` and `entitlements`. This fix allows customized json payload so the library is compatible

Add any other context about the problem here. For example:
 - Java version:13
 - SCIM 2 SDK version: [e.g. 2.3.4, master branch]

**Solution**
All fields in `MapperFactory` and method `createObjectMapper` should not be static. Just removing `static` solves the problem

kqarryzada:
Previously, the fields of the MapperFactory were static. This
effectively allowed only a single instance of the object, and also made
it impossible to override the createObjectMapper() method with personal
customizations like adding serializers/deserializers. The SCIM SDK
already treated MapperFactory instances as if they were not static, so
we will now permit flexible customization. The documentation makes it
clear that the superclass's object mapper must be fetched first for
optimal behavior.

Reviewer: vyhhuang
Reviewer: dougbulkley

JiraIssue: DS-49067
Resolves #147

### DIFF
--- a/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/MapperFactory.java
+++ b/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/MapperFactory.java
@@ -24,6 +24,7 @@ import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.MapperFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.SerializationFeature;
+import com.fasterxml.jackson.databind.json.JsonMapper;
 import com.fasterxml.jackson.databind.module.SimpleModule;
 import com.unboundid.scim2.common.annotations.NotNull;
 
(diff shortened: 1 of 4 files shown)


scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/MapperFactory.java
```diff
@@ -152,9 +221,20 @@ public MapperFactory setSerializationCustomFeatures(
...
   @NotNull
-  public static ObjectMapper createObjectMapper()
+  public ObjectMapper createObjectMapper()
   {
-    ObjectMapper mapper = new ObjectMapper(new ScimJsonFactory());
+    // Create a new object mapper with case-insensitive settings.
+    var objectMapperBuilder = JsonMapper.builder(new ScimJsonFactory());
+
+    // Do not care about case when de-serializing POJOs.
+    objectMapperBuilder.enable(ACCEPT_CASE_INSENSITIVE_PROPERTIES);
```
kqarryzada, line 230: The original mapper.configure() call for this property (which I've now deleted) was marked as deprecated by Jackson, so I've updated the logic to instead set the property on an object mapper builder instead. I'm not sure why other setting types (e.g., deserialization features) are okay with setting this directly on an object mapper, but it seems to be what the library encourages.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/ScimJsonFactory.java
```diff
@@ -68,7 +69,8 @@ protected ScimJsonFactory(@NotNull final ScimJsonFactory sourceFactory,
   JsonParser createScimFilterParser(@NotNull final Reader r)
       throws IOException
   {
-    IOContext ctxt = _createContext(r, false);
+    ContentReference reference = ContentReference.construct(true, r);
+    IOContext ctxt = _createContext(reference, false);
```
kqarryzada, line 73: This is an unrelated change that resolves another deprecation warning. I've kept this in a separate commit.

vyhhuang (approved):

dougbulkley (approved):

PR #234, opened 2024-11-01
Overhaul GenericScimResource documentation.

kqarryzada:
Updated the documentation for GenericScimResource objects to provide clearer messaging and better examples on how the class can help define a new resource type, as well as how clients can leverage the library methods effectively. The Javadocs have been standardized for consistent messaging.

This change includes some test updates to validate some of the behavior that was documented previously, but did not have coverage.

Reviewer: vyhhuang
Reviewer: dougbulkley

JiraIssue: DS-49314

### DIFF
--- a/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/GenericScimResource.java
+++ b/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/GenericScimResource.java
@@ -46,22 +46,49 @@ import java.util.List;
 import java.util.Objects;
 
 /**
- * <p>A generic SCIM object.  This object can be used if you have no
- * Java object representing the SCIM object being returned.</p>
+ * A generic SCIM object. This class can be used if you have no Java object
+ * representing the SCIM object being returned.
+ * <br><br>
+ * This object can be used when the exact structure of the SCIM object, which
+ * will be received as JSON text, is not known. This class provides methods that
+ * can read attributes from those objects without needing to know the schema
+ * ahead of time. If, however, the SCIM object you are working with is clearly
+ * defined and has an established structure, you could still use this object,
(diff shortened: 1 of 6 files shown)


dougbulkley (approved):
I did not take the time to review all your documentation changes, but I did look through your test additions thoroughly. 

I commend you for taking time to make such good contributions to the documentation for the sdk users.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/GenericScimResource.java
```diff
@@ -688,36 +648,36 @@ public GenericScimResource addStringValues(@NotNull final String path,
...
+   *   gsr.addStringValues("starRealmsCards", "BattleCruiser);
+   *
+   *   // This method is useful when the new values are already in a list.
+   *   gsr.addStringValues("starRealmsCards", List.of("BattleCruiser"));
    *
-   *   gsr.addStringValues(Path.fromString("path2"), path2values)
-   *   where path2values is a List of String.
-   *   Would create a new array called "path2"
-   *   <p>
+   *   // Creates a new array of values on the resource.
```
vyhhuang, line 670: It might be worth expanding this comment (and the matching ones for the other data types) to something like 
```
// Creates a new array of values called "sevenWondersCard" on the resource.
```

vyhhuang (approved):

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/GenericScimResource.java
kqarryzada: I considered this, but this makes the sentence a bit wordier and I was worried that sentences like that might require a re-read to understand. My hope is that the code makes it clear what the key name will be in the JSON.

PR #235, opened 2024-11-15
Allow deserializing null Resources in some cases

### LINKED ISSUE #233: The ListResponse's Resources attribute should be conditionally required
**Describe the bug**
Considering the RFC specification, the Resources field of ListResponse is required only if the "totalResults" is non-zero (https://datatracker.ietf.org/doc/html/rfc7644#section-3.4.2). In the current implementation (https://github.com/pingidentity/scim2/blob/master/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/messages/ListResponse.java#L74), the "Resources" attribute is marked as required, and if a response with no resources is received, a deserialization exception is thrown.

**To Reproduce**
Deserialize a ListResponse with no resources.

**Expected behavior**
The Resources attribute is to be required only if the "totalResults" is non-zero, or to be marked as optional and the validation is to be expected from the consumer of the class.

**Additional context**
 - Java version: 17
 - SCIM 2 SDK version: 2.3.7

kqarryzada:
RFC 7644 states that the Resources array is permitted to be null when the value of totalResults is 0. Previously, the SCIM SDK would throw an exception on any attempt to convert a JSON string to a ListResponse if the Resources array was missing.

The SDK has been updated to successfully process JSON strings with a null array, as long as the totalResults or itemsPerPage fields are explicitly set to 0. The latter is not technically requested in the specification. but it seems probable that a service provider could return a null array in this case even though it is not correct. There is arguably little benefit to rejecting the case where itemsPerPage is 0 and the Resources array is null, so we will be permissive of this to prevent further deserialization errors, as they can be difficult for application developers to work around.

Note that this commit does not change any behavior for serializing a ListResponse object to a JSON string. By default, the SCIM SDK will always print this case as an explicit empty array.

Reviewer: vyhhuang
Reviewer: dougbulkley

JiraIssue: DS-49416

Resolves #233

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,15 @@ can be used to define resource types for objects that don't have a strongly defi
 class-level Javadoc describes how to interface with the object effectively, and the methods now
 provide clearer examples of how they can be used.
 
+Updated the class-level documentation of `ListResponse` to provide more background about the
+resource type and how to interface with the object.
+
+Updated the `ListResponse` class to prevent deserialization errors when the `Resources` array is
+`null`. This is now permitted when `totalResults` and/or `itemsPerPage` is set to 0.
+[RFC 7644 Section 3.4.2](https://datatracker.ietf.org/doc/html/rfc7644#section-3.4.2)
+explicitly states that the `Resources` array may be null when `totalResults` is 0, so the SCIM SDK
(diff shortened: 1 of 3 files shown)


dougbulkley (approved):
Looks good to me.

vyhhuang (approved):

PR #238, opened 2025-01-30
Update default behavior for PATCH ADDs with filter

kqarryzada:
In the 3.2.0 release, we updated the SCIM SDK so that add operations with a filter could target existing values on the resource via a configurable property on the PatchOperation class. This is more appropriate than the previous behavior, but was not made the default.

This commit updates the SDK to target existing values by default. The project version has been updated to 4.0.0-SNAPSHOT signify this change.

Reviewer: vyhhuang
Reviewer: dougbulkley

JiraIssue: DS-49617

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,15 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](https://semver.org/).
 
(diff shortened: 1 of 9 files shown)


dougbulkley (approved):

vyhhuang (approved):

PR #245, opened 2025-03-19
Upgrade to Java 17

kqarryzada:
In the interest of maintaining compatibility with newer versions of
Java, the SCIM SDK is now compiled with Java 17. This change is
primarily motivated by our dependencies making it harder to support Java
11. This commit includes dependency version bumps.

Reviewer: vyhhuang
Reviewer: dougbulkley

JiraIssue: DS-49655

More background and testing details are available on Jira.

### DIFF
--- a/scim2-sdk-common/src/test/java/com/unboundid/scim2/common/FilterEvaluatorTestCase.java
+++ b/scim2-sdk-common/src/test/java/com/unboundid/scim2/common/FilterEvaluatorTestCase.java
@@ -56,70 +56,69 @@ public class FilterEvaluatorTestCase
     date = new Date();
 
     node = JsonUtils.getObjectReader().
-        readTree("{\n" +
-            "    \"externalId\": \"user:externalId\",\n" +
-            "    \"id\": \"user:id\",\n" +
-            "    \"meta\": {\n" +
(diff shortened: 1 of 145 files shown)


scim2-sdk-common/src/test/java/com/unboundid/scim2/common/FilterEvaluatorTestCase.java
```diff
@@ -399,6 +398,25 @@ public void testCombiningFilterEquals() throws Exception
     );
     assertThat(orFilter.equals(orFilterDifferentOrder)).isTrue();
     assertThat(orFilterDifferentOrder.equals(orFilter)).isTrue();
+
+    // Test nested AND/OR filters.
+    Filter filter1 = Filter.fromString(
```
kqarryzada, line 403: I incorporated this test case since it exposes a regression that occurred with an earlier version of this update. It seemed useful to include that in our own unit tests.

vyhhuang (approved):

dougbulkley (approved):
Khalid to provide me builds of the product with these sdk changes so I can test using our ETAH automation. I will post the results of that in the associated JIRA>

PR #250, opened 2025-04-30
Allow extending scim2-sdk-common model classes

### LINKED ISSUE #81: Ignoring unknown fields?

kqarryzada:
The SCIM 2 SDK model classes were originally intended to help build
services and clients that closely follow the rules of the
specification. To enforce this, many core model classes are final, with
the perspective that they should never to be modified.

Unfortunately, this does not work well if a client is interfacing with a
service provider that does not adhere to the standard, such as
containing extra fields outside of an extension schema. If those
additional fields contain necessary data, then extending the model
classes would be useful so that an application can capture that data.

To help SDK clients with this problem, several common model classes
(e.g., ErrorResponse) have been updated to be non-final. This commit
also contains a change that makes it possible to use the builder
pattern with the Meta class.

Reviewer: vyhhuang
Reviewer: dougbulkley

JiraIssue: DS-18005

Resolves #81


dougbulkley (approved):

vyhhuang (approved):

PR #253, opened 2025-05-22
Remove the deprecated ScimDateFormat class.

kqarryzada:
ScimDateFormat was the legacy approach for converting ISO 8601 timestamps between JSON string values and Java objects. Since the 2.2.2 release, we have used DateTimeUtils for this instead, since Jackson deprecated their ISO8601Utils class. ScimDateFormat has been deprecated for several years now, so it is a good time to remove it before the next major release.

This commit also updates the Javadoc AtclauseOrder fields since we used to have an exception for ScimDateFormat.

Reviewer: vyhhuang
Reviewer: dougbulkley

JiraIssue: DS-50088


vyhhuang (approved):

kqarryzada:
Added a correction to the release note and commit message, as ScimDateFormat was in fact deprecated in 2.2.2, but DateTimeUtils was released in 2.3.0.

PR #254, opened 2025-06-03
Clean up POM and remove Guava test dependency.

kqarryzada:
The Google Guava dependency is no longer used in any of the test code, since the Java 17 upgrade allowed us to migrate such usages to methods that are now supported natively. We don't anticipate that this dependency will be needed again in the future, so it is being removed.

This commit also includes some POM cleanup, and resolves an issue where an older version of plugins were being run during build time.

Reviewer: vyhhuang
Reviewer: dougbulkley

JiraIssue: DS-50007


vyhhuang (approved):

PR #255, opened 2025-06-06
Make GenericScimResource extendable.

kqarryzada:
The 4.0.0 release will have several classes that are no longer final. One object that should be included is GenericScimResource. Much like how custom resource types can be created today by extending BaseScimResource, this change allows library clients to create their own resource types that extend GenericScimResource with a dedicated class name.


vyhhuang (approved):

PR #257, opened 2025-07-18
Simplify Jackson calls to fix deprecation warnings

kqarryzada:
ObjectNode.fields() is now deprecated in Jackson 2.19. As a result, the SDK has been updated to use the ObjectNode.properties() method instead, which is friendlier to iteration in most cases.

Evaluation of objects and arrays has also been updated to use Java pattern variables, as this provides an easy (and cleaner) way to obtain an ObjectNode or ArrayNode reference.

Reviewer: vyhhuang
Reviewer: dougbulkley

JiraIssue: DS-50361

### DIFF
--- a/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/messages/PatchOperation.java
+++ b/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/messages/PatchOperation.java
@@ -418,7 +418,7 @@ public abstract class PatchOperation
         // prepare to add this value ourselves.
         jsonAttribute = JsonUtils.getJsonNodeFactory().arrayNode(1);
       }
-      if (!jsonAttribute.isArray())
+      if (!(jsonAttribute instanceof ArrayNode attr))
       {
         throw BadRequestException.invalidSyntax(
             "The patch operation could not be processed because a complex"
@@ -426,7 +426,6 @@ public abstract class PatchOperation
                 + "' is single-valued"
         );
       }
-      ArrayNode attribute = (ArrayNode) jsonAttribute;
 
       // When operations with a value filter add data, we can either append
       // the data to a new value in the multi-valued attribute, or we can update
@@ -440,7 +439,7 @@ public abstract class PatchOperation
       ObjectNode matchedValue = null;
       if (!APPEND_NEW_PATCH_VALUES_PROPERTY)
       {
-        matchedValue = fetchExistingValue(attribute, valueFilter, attributeName);
+        matchedValue = fetchExistingValue(attr, valueFilter, attributeName);
       }
 
       // If there are no existing values that match the filter, or if no values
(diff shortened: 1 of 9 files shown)


scim2-sdk-common/src/main/java/com/unboundid/scim2/common/messages/PatchOperation.java
```diff
@@ -440,7 +439,7 @@ private void applyAddWithValueFilter(
       ObjectNode matchedValue = null;
       if (!APPEND_NEW_PATCH_VALUES_PROPERTY)
       {
-        matchedValue = fetchExistingValue(attribute, valueFilter, attributeName);
+        matchedValue = fetchExistingValue(attr, valueFilter, attributeName);
```
kqarryzada, line 442: I mostly renamed this variable because I didn't notice this line exceeds 80 characters when I first wrote it.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/JsonDiff.java
```diff
@@ -251,81 +250,77 @@ private void computeArrayNodeDiffs(
 
       // Explicitly clear all attribute values.
       operations.add(PatchOperation.remove(path));
+      return;
     }
-    else
+
+    // Go through each value and try to individually patch them first
```
kqarryzada, line 256: The changes below this look messy, but it was an effort to remove multiple layers of nesting. Most lines are not changed, with the exception of switching to `instanceof`.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/JsonUtils.java
```diff
@@ -74,7 +74,7 @@ public abstract static class NodeVisitor
     @NotNull
     abstract JsonNode visitInnerNode(@NotNull final ObjectNode parent,
                                      @Nullable final String field,
-                                     @NotNull final Filter valueFilter)
+                                     @Nullable final Filter valueFilter)
```
kqarryzada, line 77: This annotation was marked incorrectly before.

scim2-sdk-server/src/main/java/com/unboundid/scim2/server/utils/SchemaChecker.java
```diff
@@ -634,9 +631,8 @@ private void checkPartialResource(
       final boolean isPartialAdd)
           throws ScimException
   {
-
-    Iterator<Map.Entry<String, JsonNode>> i = objectNode.fields();
-    while (i.hasNext())
+    Iterator<Map.Entry<String, JsonNode>> i;
+    for (i = objectNode.properties().iterator(); i.hasNext();)
```
kqarryzada, line 635: These usages need the iterator, since they are removing nodes with `i.remove()`.

dougbulkley (approved):
Are these changes covered by existing tests? If so, can you enumerate which ones?
Also, sometimes you include a note in the changelog file (ex: previously when removing the guava dependency), why not here?

PR #258, opened 2025-07-25
Use GenericScimResource for JSON payloads

kqarryzada:
Test automation recently found that sending requests from the client library had the potential to fail if key properties are not enabled within the Java project using the SCIM SDK. Unfortunately, these properties can also be difficult to determine and fix.

To address this problem, the request builders have been updated to use GenericScimResources for JSON payloads. This class has its own custom serializer defined, which simplifies converting any ScimResource-based object into the correct JSON data.

Reviewer: vyhhuang
Reviewer: dougbulkley

JiraIssue: DS-50376


scim2-sdk-client/src/main/java/com/unboundid/scim2/client/requests/RequestBuilder.java
```diff
@@ -275,4 +278,40 @@ protected Invocation.Builder buildRequest()
...
+   * The behavior of the SCIM SDK's object mapper can be customized by updating
+   * fields in the {@link com.unboundid.scim2.common.utils.MapperFactory} class.
+   *
+   * @param resource  The SCIM resource that will be serialized into JSON.
+   * @return  An equivalent GenericScimResource.
+   *
+   * @since 4.0.1
+   */
+  @NotNull
+  protected GenericScimResource generify(@NotNull final ScimResource resource)
```
kqarryzada, line 313: I made a wrapper method for this so that there was a centralized spot that documents why a GenericScimResource is used.

vyhhuang (approved):
These changes look good to me. I appreciate the detailed comment on the `generify`method.

PR #259, opened 2025-07-28
feat: add maxResults param in SimpleSearchResults

Ripolin:
Add a param `maxResults` to limit the number of elements returns when using a `SimpleSearchResults` object (see com.unboundid.scim2.common.types.FilterConfig#maxResults).

When using `null` value , no limitation is used.

### DIFF
--- a/scim2-sdk-server/src/main/java/com/unboundid/scim2/server/utils/SimpleSearchResults.java
+++ b/scim2-sdk-server/src/main/java/com/unboundid/scim2/server/utils/SimpleSearchResults.java
@@ -73,12 +73,13 @@ public class SimpleSearchResults<T extends ScimResource>
    *
    * @param resourceType The resource type definition of result resources.
    * @param uriInfo The UriInfo from the search operation.
+   * @param maxResults The maximum number of resources returned in a response.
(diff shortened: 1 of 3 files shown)


kqarryzada:
Hi @Ripolin. Unfortunately, Ping Identity currently does not accept third-party code contributions for this repository (see CONTRIBUTING.md). In general, I would suggest opening a GitHub issue with some context for what you're looking for. However, in this case, I think we'd be unlikely to incorporate code changes to support behavior similar to what you're trying to achieve.

The main issue to note is that the SimpleSearchResults object is used to represent a barebones response implementation. From the Javadoc:

> A utility ListResponseStreamingOutput... for simple search implementations that always returns the entire result set.

We would not want to add complexity to this implementation. If you're seeking more specific behavior, I would suggest creating your own subclass of ListResponseStreamingOutput (or even SimpleSearchResults) to achieve this. Alternatively, on the endpoint where you are processing this request, you could potentially use a `jakarta.ws.rs.DefaultValue` annotation to obtain a default value for the `count` query parameter which already does this. However, I haven't done that before so I don't know if it's what you need.

PR #260, opened 2025-08-08
Enhance the Path class with new methods and docs

kqarryzada:
Added two new methods, Path.of() and Path.getLastElement(), to simplify interactions with the Path class. In particular, this aims to make the creation of hard-coded simple attributes less verbose. As part of this update, the class-level Javadoc has been updated with more background and examples on how to interface with the class.

Reviewer: vyhhuang
Reviewer: dougbulkley

JiraIssue: DS-50434

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,17 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](https://semver.org/).
 
 ## v4.0.1 - TBD
+Added new methods to the Path class to simplify certain usages and make interaction, especially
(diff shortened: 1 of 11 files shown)


kqarryzada:
Thanks @dougbulkley. I've added some minor doc changes to make it clearer that this is an alternate method, and not the primary. That way, if there's confusion, the better option is chosen as the default. 

I envision this mostly being useful as a way to pass path objects into filters and patch operations in a more concise manner. Using `fromString` requires exception handling that is not necessary when you're passing in a hard-coded string like "userName" that is known to be valid.

PR #266, opened 2025-09-19
Deprecate BaseScimResource#addExtensionValue

kqarryzada:
This commit deprecates the addExtensionValue methods, which would add schema extension data in array form. This formats the SCIM data in a way that is also incompatible with deserialization. To avoid confusion about whether such data formats are permitted, these methods will be removed in a future release.

Reviewer: vyhhuang
Reviewer: dougbulkley

JiraIssue: DS-50646


vyhhuang (approved):

PR #267, opened 2025-09-26
Support non-standard patch group membership remove

### LINKED ISSUE #174: No support for removing group members
Patchoperation with opType remove does not save the objectNode value .
This makes it hard to support removing group members according to this:
https://help.salesforce.com/s/articleView?id=sf.identity_scim_manage_groups.htm&type=5.

I want to use PatchRequest and PatchOperation as pojos and then handle the deleting in my server, when removing the objectNode, this is not possible.

kqarryzada:
Some SCIM service providers mandate patch remove requests that violate the standard's definition. This has been seen specifically for requests that seek to update a group to remove one or more members. For more details and background, see the new PatchOperation#removeOpValue method, as well as the documentation for the
NonStandardRemoveOperationTest test class.

To help clients interface with these endpoints, a new removeOpValue() method has been added that allows setting a value on a remove operation conditionally. Support for applying these updates to a SCIM resource has also been added to provide server-side support.

Reviewer: vyhhuang
Reviewer: dougbulkley

JiraIssue: DS-50483
Resolves #174

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](https://semver.org/).
 
-## v4.0.1 - TBD
+## v4.1.0 - TBD
 Added new methods to the Path class to simplify certain usages and make interaction, especially
 instantiation, less verbose. These include:
 * Creation of simple attributes (e.g., `username`) previously had to be performed with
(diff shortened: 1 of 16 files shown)


dougbulkley (approved):
Thorough testing as usual, thank you.

vyhhuang (approved):
The changes look fine to me. I do wonder if it would be preferable to use `setRemoveOpValue()` instead of `removeOpValue()` though -- while the former is wordier, it does eliminate possible ambiguity around what the method does.

kqarryzada:
@vyhhuang, I agree about the method name. I've updated it as you've suggested.

PR #270, opened 2025-10-15
Update dependencies to the latest versions. 

kqarryzada:
Updated dependencies and plugins to the latest releases. This includes
changes that anticipate an upgrade to the newly released Jackson 3.0.0.

Jackson 3.x contains many backwards-incompatible changes that are not
possible to implement unless clients update their own code, such as
changing package names from "com.fasterxml" to "tools.jackson". There
are a small number of API usages that will be incompatible with Jackson
3.x, but already have alternative approaches available in 2.20. Thus,
this commit includes some changes to use alternate Jackson APIs, though
these changes are likely not exhaustive.

Reviewer: vyhhuang
Reviewer: dougbulkley

JiraIssue: DS-50800

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](https://semver.org/).
 
 ## v4.1.1 - TBD
+Updated the following dependencies:
+* Jackson: 2.19.2 -> 2.20.0
+* Jersey: 3.1.10 -> 3.1.11
 
 ## v4.1.0 - 2025-Oct-06
 Added new methods to the Path class to simplify certain usages and make interaction, especially


scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/FilterEvaluator.java
```diff
@@ -444,22 +443,20 @@ private Iterable<JsonNode> getCandidateNodes(@NotNull final Path path,
...
-      }
-      return true;
+      return node.isNull();
     }
-    else
+
+    for (JsonNode jsonNode : node)
     {
-      return node.isNull();
+      if (!isEmpty(jsonNode))
```
vyhhuang: Honestly, the original version of this logic is a little baffling to parse, but as far as I can tell this is logically identical so it's fine.

PR #279, opened 2025-12-04
Validate non-standard patch removes on creation

kqarryzada:
To ensure better parity with older releases, the paths of non-standard remove operations are validated during object creation. Previously, validation occurred when applying/processing the operations.

Reviewer: vyhhuang
Reviewer: dougbulkley

JiraIssue: DS-50911


vyhhuang (approved):

PR #280, opened 2025-12-04
Support cursor-based pagination (RFC 9865)

kqarryzada:
RFC 9865 is an update to the SCIM 2 standard that defines cursor pagination, an alternative to the index-based API pagination defined in RFC 7644. This allows SCIM services to offer an alternative means for paginating list responses if index-based pagination is not feasible to implement. For more background on cursor-based pagination, see the documentation for the new PaginationConfig class.

The SCIM SDK has been updated to support all newly defined parameters in this standard update, including new exception types. This includes updates to the client, server, and common libraries. These changes are designed to preserve backward compatibility for SCIM applications that do not need cursor-based pagination, so the SDK is still compatible with clients and services that do not yet support RFC 9865.

Reviewer: vyhhuang
Reviewer: dougbulkley

JiraIssue: DS-50973

### DIFF
--- a/scim2-sdk-common/src/test/java/com/unboundid/scim2/common/SearchRequestTestCase.java
+++ b/scim2-sdk-common/src/test/java/com/unboundid/scim2/common/SearchRequestTestCase.java
@@ -108,7 +108,6 @@ public class SearchRequestTestCase
           "filter": "displayName sw \\"smith\\"",
           "sortBy": "id",
           "sortOrder": "descending",
-          "startIndex": 21,
           "cursor": "YkU3OF86Pz0rGv",
           "count": 5
         }""";
@@ -124,7 +123,7 @@ public class SearchRequestTestCase
         .isEqualTo(Filter.sw("displayName", "smith").toString());
     assertThat(cursorRequest.getSortBy()).isEqualTo("id");
     assertThat(cursorRequest.getSortOrder()).isEqualTo(SortOrder.DESCENDING);
-    assertThat(cursorRequest.getStartIndex()).isEqualTo(21);
+    assertThat(cursorRequest.getStartIndex()).isNull();
     assertThat(cursorRequest.getCursor()).isEqualTo("YkU3OF86Pz0rGv");
     assertThat(cursorRequest.getCount()).isEqualTo(5);
 


scim2-sdk-client/src/main/java/com/unboundid/scim2/client/requests/SearchRequestBuilder.java
```diff
@@ -163,6 +174,63 @@ public SearchRequestBuilder page(final int startIndex,
...
+   *   GET /Users?cursor=&amp;count=10
+   * </pre>
+   *
+   * @param count   The desired maximum number of query results per page.
+   * @return This builder.
+   *
+   * @since 5.0.0
+   */
+  @NotNull
+  public SearchRequestBuilder firstPageCursorWithCount(final int count)
```
kqarryzada: I included "WithCount" because a call like `firstPageCursor(10)` seemed very confusing.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/messages/ListResponse.java
```diff
@@ -180,30 +252,91 @@ public class ListResponse<T> extends BaseScimResource
...
+   * @param itemsPerPage  The number of resources returned in the list response
+   *                      page.
+   *
+   * @throws IllegalStateException If the {@code resources} list is {@code null}
+   *                               and {@code totalResults} is non-zero.
+   */
+  public ListResponse(final int totalResults,
+                      @NotNull final List<T> resources,
+                      @Nullable final Integer startIndex,
+                      @Nullable final Integer itemsPerPage)
```
kqarryzada: This constructor matches the old `@JsonCreator` one for backward compatibility.

scim2-sdk-common/src/test/java/com/unboundid/scim2/common/SearchRequestTestCase.java
```diff
@@ -32,83 +32,160 @@
...
-   * @throws ScimException If an error occurs.
+   * Verifies that serialization and deserialization of search request objects
+   * result in an expected structure.
    */
   @Test
-  public void testSearchRequest() throws Exception
+  public void testSerialization() throws Exception
   {
-    // Test case insensitivity
-    SearchRequest searchRequest = JsonUtils.getObjectReader().
```
kqarryzada: I added my new test case and realized it also encompassed some behavior tested by this existing test. I removed the old test since the newer test matches other serialization tests better.

PR #290, opened 2026-04-10
Update README and public docs.

kqarryzada:
Updated the documentation with recent information. The README now
provides additional details and background, such as highlighting recent
support for features like cursor-based pagination.

Reviewer: dougbulkley
Reviewer: vyhhuang

JiraIssue: DS-51373

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,8 +24,8 @@ different SCIM services can have variance by enforcing their own constraints, bu
 common ground.
 
 ## Advantages of the UnboundID SCIM 2 SDK
-The UnboundID SCIM 2 SDK provides many strong benefits for applications that need support for
-SCIM 2.0:
+The UnboundID SCIM 2 SDK provides many strong benefits for applications that need to communicate
+with SCIM 2.0 clients or servers:
 
 * Full support for the SCIM 2.0 protocol as defined by the latest specification. This includes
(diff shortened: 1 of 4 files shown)


dougbulkley (approved):

README.md
```diff
@@ -63,74 +109,101 @@ import com.unboundid.scim2.common.GenericScimResource;
...
-  .setType("home")
-  .setPrimary(true)
-  .setValue("babs@example.com");
-user.setEmails(email);
+// Create a user with a POST request by calling create().
+UserResource user = new UserResource()
+    .setUserName("babs")
+    .setPassword("secret");
+user.setName(new Name()
+    .setGivenName("Barbara")
```
vyhhuang, line 123: Nit: I think this (and the `new Email().setType...` section below) could either be given a larger indent or put onto a single line. For a second I thought a trailing parenthesis was needed.

vyhhuang (approved):
I had a minor comment, but these changes look good to me.

README.md
kqarryzada: I moved the closing parentheses for both of these onto the next line, which should hopefully make it more clear. Do you think that's sufficient, or would you prefer something like a single line instead?

vyhhuang: That's fine by me 👍

PR #292, opened 2026-04-17
Add new applyToResource() patch update methods.

kqarryzada:
The existing PatchOperation.apply() and PatchRequest.apply() methods
have provided a viable way to process patch updates since the inception
of the project. However, they require that clients manually perform
conversions between objects and JsonNodes. Applications would also
frequently instantiate and maintain their own object mapper for this
(instead of using JsonUtils), adding further friction.

To make this process smoother, new applyToResource() methods have been
added. These accept any ScimResource-based object and returns the same
instance type in a single line of code. This removes the need for manual
JSON conversions, and provides a streamlined and more direct manner for
applying SCIM patch updates.

Existing helper methods that mimicked this behavior in the unit tests
have been updated to leverage the new API. The new methods have also
been added to the documentation as the recommended approach for update
processing.

Reviewer: dougbulkley
Reviewer: vyhhuang

JiraIssue: DS-51406

### DIFF
--- a/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/messages/PatchRequest.java
+++ b/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/messages/PatchRequest.java
@@ -34,14 +34,18 @@ package com.unboundid.scim2.common.messages;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.unboundid.scim2.common.ScimResource;
 import com.unboundid.scim2.common.annotations.NotNull;
 import com.unboundid.scim2.common.annotations.Nullable;
 import com.unboundid.scim2.common.annotations.Schema;
 import com.unboundid.scim2.common.annotations.Attribute;
+import com.unboundid.scim2.common.exceptions.BadRequestException;
 import com.unboundid.scim2.common.exceptions.ScimException;
 import com.unboundid.scim2.common.BaseScimResource;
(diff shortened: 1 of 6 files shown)


scim2-sdk-common/src/main/java/com/unboundid/scim2/common/messages/PatchRequest.java
```diff
@@ -148,19 +164,59 @@ public Iterator<PatchOperation> iterator()
...
+   */
+  @NotNull
+  public <T extends ScimResource> T applyToResource(@NotNull final T resource)
+      throws ScimException
+  {
+    try
+    {
+      // Use this instead of PatchOperation#applyToResource so that the Jackson
+      // parsing and conversion is only done once.
+      GenericScimResource updatedJson = apply(resource.asGenericScimResource());
```
kqarryzada, line 210: I ran a rudimentary evaluation of this by taking one of the new unit test cases and repeatedly applying the given patch request 100K times. The method used here completed in 350 ms, but when I updated this method to instead iterate like the following, it took around 720 ms:

```java
T updatedResource = resource;
for (PatchOperation op : this)
{
  updatedResource = op.applyToResource(updatedResource);
}
```

This makes sense, since the patch operation has to repeatedly convert the object between JSON and a POJO instead of just doing this process once. We potentially lose out on a more descriptive error, but I think this tradeoff is worth it because the other approach is pretty wasteful.

dougbulkley (approved):

vyhhuang (approved):

PR #294, opened 2026-04-24
Fix import statement order.

kqarryzada:
In the upcoming Jackson 3 migration, import statements in the project are going to be slightly reordered since namespaces are changing from "com.fasterxml.jackson" to "tools.jackson". Early validation has revealed that there are some places where import statements are not in alphabetical order.

To keep consistency between the future releases and 5.1.0, this commit makes import statement ordering consistent and supplies a Checkstyle rule to enforce this. Test files have been updated, but the Checkstyle rule only applies to the project files, as we care about consistency there much more than in the unit tests.

Reviewer: vyhhuang


config/scim2-parent-checkstyle.xml
```diff
@@ -114,6 +114,20 @@
     <module name="UnusedImports" />
 
 
+    <!-- Sort imports in alphabetical order. Use IntelliJ's style. -->
```
kqarryzada, line 117: I pulled this from the Checkstyle documentation: https://checkstyle.org/checks/imports/importorder.html#Example4-config

vyhhuang (approved):

PR #295, opened 2026-04-24
Add system property for UTC timestamps. 

kqarryzada:
In previous releases where JAX-B was used for timestamp deserialization,
all timestamps that occurred in the default region were placed in a
timezone with an ID of "GMT+00:00". In many cases, it may be desirable
to use UTC as the default value instead.

This commit adds a new system property that can be enabled so that
deserialized timestamps will result in Calendar objects with a timezone
corresponding to UTC. This also updates existing customizable
properties to use Java system properties as well, without changing
their default value.

Reviewer: vyhhuang
Reviewer: dougbulkley

JiraIssue: DS-51419

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,14 @@ UserResource updatedUser = patchRequest.applyToResource(user);
 ```
 The existing `apply()` methods are not deprecated and may still be used.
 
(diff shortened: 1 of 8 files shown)


vyhhuang (approved):

dougbulkley (approved):

PR #296, opened 2026-04-24
Deprecate parameters in anticipation of Jackson 3.

kqarryzada:
These changes primarily focus on the MapperFactory class. In Jackson 3.x, object mappers will be read only instances. Furthermore, some class names such as "JsonParser.Feature" have been changed, so our existing methods will require upkeep to facilitate customizations.

To simplify configuration management, the SCIM SDK now exposes the builder object that is used to create new object mapper instances. This stored builder instance provides more flexibility for handling configuration updates, and is an improvement over manually tracking each Jackson configuration type in our MapperFactory class.

This commit also deprecates the PATCH annotation, as JAX-RS has provided an annotation for this type for many releases now. This will be removed with the upcoming Jackson 3 upgrade.

Reviewer: dougbulkley
Reviewer: vyhhuang

JiraIssue: DS-51420

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,17 @@ have been updated so that they can be toggled by a Java system property as well.
 changing these values without requiring a code change. Note that if the system property value is
 changed at runtime, a restart is required for the change to take effect.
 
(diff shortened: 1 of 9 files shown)


vyhhuang (approved):

dougbulkley (approved):

PR #297, opened 2026-05-01
Add convenience methods to Meta.java

kqarryzada:
This commit updates the Meta class with convenience methods that
simplify the interface for interaction with other data types. It is now
possible to provide UNIX timestamps for created/lastModified
timestamps, and location values may be set or fetched as strings.

This PR also includes dependency updates.

Reviewer: vyhhuang

JiraIssue: DS-51461

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -84,13 +84,13 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <compileSource>17</compileSource>
     <main.basedir>${project.basedir}</main.basedir>
-    <jackson.version>2.21.2</jackson.version>
+    <jackson.version>2.21.3</jackson.version>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -327,9 +330,7 @@
         <configuration>
           <tagNameFormat>scim-@{project.version}</tagNameFormat>
           <goals>deploy</goals>
-          <useReleaseProfile>false</useReleaseProfile>
           <checkModificationExcludeList>pom.xml</checkModificationExcludeList>
-          <waitBeforeTagging>3</waitBeforeTagging>
```
kqarryzada, line 332: `useReleaseProfile` is false by default, and the docs for `waitBeforeTagging` mention that it is useful for SVN repositories, so it's a legacy value that is not needed.

vyhhuang (approved):

PR #299, opened 2026-05-21
Fix deserialization of GenericScimResource.

kqarryzada:
Jackson 3 enabled the DeserializationFeature.FAIL_ON_TRAILING_TOKENS setting by default. This caused problems for the custom deserializer since extra JSON data can be present if a GenericScimResource is embedded within a list response. As a result, this commit updates the deserializer to address this edge case.

This also includes a nullability annotation correction and minor test cleanup.

Reviewer: dougbulkley
Reviewer: vyhhuang

JiraIssue: DS-51507

### DIFF
--- a/scim2-sdk-common/src/test/java/com/unboundid/scim2/common/GenericScimResourceTest.java
+++ b/scim2-sdk-common/src/test/java/com/unboundid/scim2/common/GenericScimResourceTest.java
@@ -33,6 +33,9 @@
 package com.unboundid.scim2.common;
 
 import com.unboundid.scim2.common.exceptions.ScimException;
+import com.unboundid.scim2.common.exceptions.runtime.ScimDeserializeException;
+import com.unboundid.scim2.common.messages.ListResponse;
+import com.unboundid.scim2.common.types.GroupResource;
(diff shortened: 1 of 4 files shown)


dougbulkley (approved):

vyhhuang (approved):

scim2-sdk-common/src/test/java/com/unboundid/scim2/common/GenericScimResourceTest.java
```diff
@@ -130,13 +118,6 @@ public void testBasicParsing() throws Exception
     Meta meta = cso.getMeta();
     Assert.assertNotNull(meta);
 
-    Assert.assertEquals(
-        dateFormat.format(meta.getCreated().getTime()),
-        "2015-02-27T11:28:39.042Z");
```
kqarryzada, line 135: I forgot to mention, but I removed these since I originally swapped it with DateTimeUtils.format(). That was better than using a manual formatter that only supports UTC. However, I realized that the next case using ObjectNode already implicitly calls into this. Thus, I removed this so it wouldn't be duplicitous.

PR #301, opened 2026-06-16
Add a new schema checker option

kolehey:
Added a new ALLOW_OBJECT_VALUED_STRING_SUB_ATTRIBUTES schema checker option to allow string-type sub-attributes whose value is a json object.

Reviewer: kqarryzada
Reviewer: dougbulkley

JiraIssue: DS-51562

### DIFF
--- a/scim2-sdk-server/src/main/java/com/unboundid/scim2/server/utils/SchemaChecker.java
+++ b/scim2-sdk-server/src/main/java/com/unboundid/scim2/server/utils/SchemaChecker.java
@@ -201,6 +201,15 @@ public class SchemaChecker
      * that are not defined by the definition of the parent attribute.
      */
     ALLOW_UNDEFINED_SUB_ATTRIBUTES,
+
+    /**
+     * Relax SCIM 2 standard schema requirements by allowing STRING-typed
+     * sub-attributes to contain JSON object values.
+     * <p>
+     * Per RFC 7643, complex sub-attributes are not supported; this option
+     * relaxes that constraint for non-conformant resources.
+     */
+    ALLOW_OBJECT_VALUED_STRING_SUB_ATTRIBUTES
   }
 
   @NotNull
@@ -1040,6 +1049,19 @@ public class SchemaChecker
     switch (attribute.getType())
     {
       case STRING:
+        if (!node.isString())
+        {
+          if (enabledOptions.contains(
+              Option.ALLOW_OBJECT_VALUED_STRING_SUB_ATTRIBUTES)
+              && node.isObject() && path.size() > 1)
+          {
+            return;
+          }
+          results.syntaxIssues.add(prefix + "Value for attribute " + path +
+              " must be a JSON string");
+          return;
+        }
+        break;
       case DATETIME:
       case REFERENCE:
         if (!node.isString())
--- a/scim2-sdk-server/src/test/java/com/unboundid/scim2/server/utils/SchemaCheckerTestCase.java
+++ b/scim2-sdk-server/src/test/java/com/unboundid/scim2/server/utils/SchemaCheckerTestCase.java
@@ -2268,6 +2268,133 @@ public class SchemaCheckerTestCase
     };
   }
 
+  /**
+   * Test case for the allow object valued string sub-attributes option.
+   *
+   * @throws Exception if an error occurs.
+   */
+  @Test
+  public void testAllowObjectValuedStringSubAttributesOption()
+      throws Exception
+  {
+    // Build a schema with a complex parent attribute whose sub-attribute is
+    // STRING-typed but stores a JSON object value.
+    AttributeDefinition.Builder builder = new AttributeDefinition.Builder();
+    builder.setName("iamSvcUsrData");
+    builder.setType(AttributeDefinition.Type.STRING);
+    builder.setMultiValued(true);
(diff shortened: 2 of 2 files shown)


scim2-sdk-server/src/main/java/com/unboundid/scim2/server/utils/SchemaChecker.java
```diff
@@ -201,6 +201,15 @@ public enum Option
...
     ALLOW_UNDEFINED_SUB_ATTRIBUTES,
+
+    /**
+     * Relax SCIM 2 standard schema requirements by allowing STRING-typed
+     * sub-attributes to contain JSON object values.
+     * <p>
+     * Per RFC 7643, complex sub-attributes are not supported; this option
+     * relaxes that constraint for non-conformant resources.
+     */
+    ALLOW_OBJECT_VALUED_STRING_SUB_ATTRIBUTES
```
kqarryzada, line 212: This should have a comma at the end.

scim2-sdk-server/src/main/java/com/unboundid/scim2/server/utils/SchemaChecker.java
```diff
@@ -1040,6 +1049,19 @@ private void checkAttributeValue(
     switch (attribute.getType())
     {
       case STRING:
+        if (!node.isString())
+        {
+          if (enabledOptions.contains(
+              Option.ALLOW_OBJECT_VALUED_STRING_SUB_ATTRIBUTES)
+              && node.isObject() && path.size() > 1)
```
kqarryzada, line 1056: The node type is checked twice, which we can avoid (along with the copied code) by rewriting this with the following. I also think we can rename the option:

```java
case STRING:
  if (node.isObject() && path.size() > 1
    && enabledOptions.contains(Option.ALLOW_OBJECT_AS_STRING_SUB_ATTR))
  {
    return;
  }

  // Fall through for string evaluation.
case DATETIME:
...
```

scim2-sdk-server/src/test/java/com/unboundid/scim2/server/utils/SchemaCheckerTestCase.java
```diff
@@ -2268,6 +2268,133 @@ private Object[][] getResultData()
...
+   * @throws Exception if an error occurs.
+   */
+  @Test
+  public void testAllowObjectValuedStringSubAttributesOption()
+      throws Exception
+  {
+    // Build a schema with a complex parent attribute whose sub-attribute is
+    // STRING-typed but stores a JSON object value.
+    AttributeDefinition.Builder builder = new AttributeDefinition.Builder();
+    builder.setName("iamSvcUsrData");
```
kqarryzada, line 2283: The old tests don't do this much, but we should leverage the builder pattern on this:
```java
AttributeDefinition.Builder builder = new AttributeDefinition.Builder()
    .setName(...)
    .setType(...)
```

scim2-sdk-server/src/test/java/com/unboundid/scim2/server/utils/SchemaCheckerTestCase.java
```diff
@@ -2268,6 +2268,133 @@ private Object[][] getResultData()
...
+    ObjectNode iamServiceUserJsonObjectNode = JsonUtils.getJsonNodeFactory().objectNode();
+    iamServiceUserJsonObjectNode.putArray("schemas").add("urn:id:test");
+    iamServiceUserJsonObjectNode.set("iamServiceUser", iamServiceUserNode);
+
+    SchemaChecker.Results results = refuseObjValStrSubAttributesChecker.checkCreate(
+            iamServiceUserJsonObjectNode);
+    assertEquals(results.getSyntaxIssues().size(), 1,
+        results.getSyntaxIssues().toString());
+    assertTrue(containsIssueWith(results.getSyntaxIssues(),
+        "must be a JSON string"));
```
kqarryzada, line 2339: New unit tests should use AssertJ's assertion statements instead of TestNG's. For example, this should be:

```java
assertThat(results.getSyntaxIssues()).hasSize(1);
assertThat(results.getSyntaxIssues().get(0))
    .contains("must be a JSON string");
```

For assertion statements that currently have a reason string, you can use `.as()` if you really want, but I think it's also okay to drop them from this test.

scim2-sdk-server/src/test/java/com/unboundid/scim2/server/utils/SchemaCheckerTestCase.java
```diff
@@ -2268,6 +2268,133 @@ private Object[][] getResultData()
...
+        results.getSyntaxIssues().toString());
+
+    // checkModify (PATCH replace on iamServiceUser).
+    ObjectNode currentResource = JsonUtils.getJsonNodeFactory().objectNode();
+    currentResource.putArray("schemas").add("urn:id:test");
+    currentResource.set("iamServiceUser", iamServiceUserNode);
+
+    List<PatchOperation> patchOps = Collections.singletonList(
+        PatchOperation.replace(
+            Path.root().attribute("iamServiceUser"), iamServiceUserNode));
```
kqarryzada, line 2353: If you want, this can use the string attribute directly since the method throws an exception.

kqarryzada (commented):
This looks good to me. I had some minor comments.

PR #302, opened 2026-06-18
Modernize CaseIgnoreObjectNode usages. 

kqarryzada:
This commit primarily updates CaseIgnoreObjectNode to be more aligned
with the most recent version of ObjectNode.java from Jackson. These
changes were focused on maintainability by leveraging more code re-use.
Additionally, case-insensitive evaluations have been updated to use
equalsIgnoreCase() where possible, for better performance.

This also enforces the use of a CaseIgnoreObjectNode within the
GenericScimResource class to ensure that these resources always use
case-insensitive fields. This commit uncovered a defect where the
fields would not be treated correctly if a standard ObjectNode was
provided to the constructor. This issue has now been fixed.

Reviewer: dougbulkley
Reviewer: vyhhuang

JiraIssue: DS-51630

### DIFF
--- a/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/CaseIgnoreMap.java
+++ b/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/CaseIgnoreMap.java
@@ -55,16 +55,8 @@ public class CaseIgnoreMap implements Map<String, JsonNode>
    * A wrapper around the standard String but compares and hashes them
    * in lower-case.
    */
-  private static class CaseIgnoreKey
+  private record CaseIgnoreKey(@NotNull String key)
   {
-    @NotNull
-    private final String key;
-
-    CaseIgnoreKey(@NotNull final String key)
-    {
-      this.key = key;
-    }
-
     @NotNull
     public String getKey()
     {
@@ -78,15 +70,8 @@ public class CaseIgnoreMap implements Map<String, JsonNode>
       {
         return true;
       }
-      if (o == null || getClass() != o.getClass())
-      {
-        return false;
-      }
 
-      CaseIgnoreKey that = (CaseIgnoreKey) o;
-
-      return toLowerCase(key).equals(toLowerCase(that.key));
-
+      return o instanceof CaseIgnoreKey that && key.equalsIgnoreCase(that.key);
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/CaseIgnoreMap.java
```diff
@@ -55,16 +55,8 @@ public class CaseIgnoreMap implements Map<String, JsonNode>
    * A wrapper around the standard String but compares and hashes them
    * in lower-case.
    */
-  private static class CaseIgnoreKey
+  private record CaseIgnoreKey(@NotNull String key)
   {
-    @NotNull
-    private final String key;
```
kqarryzada, line 61: I opted to use records for these internal classes to reduce boiler plate.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/CaseIgnoreObjectNode.java
```diff
@@ -65,170 +76,98 @@ public CaseIgnoreObjectNode(@NotNull final JsonNodeFactory nc)
...
-
-    return ret;
-  }
-
-  /**
-   * {@inheritDoc}
-   */
-  @Override
-  @Nullable
-  public JsonNode findValue(@NotNull final String fieldName)
```
kqarryzada, line 99: Jackson helpfully routes calls like this to the plural variant and returns the first result, so the explicit method override is not needed. It's a tradeoff between maintainability and performance (i.e., walking the whole tree), but I don't think there's that much overhead with this approach.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/CaseIgnoreObjectNode.java
```diff
@@ -65,170 +76,100 @@ public CaseIgnoreObjectNode(@NotNull final JsonNodeFactory nc)
...
-        }
-        localFoundSoFar.add(entry.getValue());
-      }
-      else
-      { // only add children if parent not added
-        localFoundSoFar = entry.getValue().findValues(fieldName, foundSoFar);
-      }
-    }
-    return localFoundSoFar;
+    List<JsonNode> parsed = findParents(propertyName, null)
```
kqarryzada, line 102: These explicitly pass null to call our method directly instead of relying on Jackson to do it.

vyhhuang (approved):

dougbulkley (approved):

PR #303, opened 2026-07-02
Update bulk patch operation handling.

kqarryzada:
The errata of RFC 7644 contains details on the intended form of bulk patch operations. It affirms that the 'data' field reflects the JSON payload that would be sent to the provided endpoint if the request were made individually to an endpoint such as 'PATCH /Users/{userID}'. Thus, it explicitly states that the JSON data provided on a bulk patch should always represent a PatchRequest resource with a "schemas" value.

This commit updates the SCIM SDK to reflect this behavior. To ensure compatibility during deserialization, extra leniency has been added when initializing bulk patch operations to reduce exceptions. After instantiation, the data field of bulk patch operation objects will always represent a full PatchRequest model.

Reviewer: dougbulkley
Reviewer: vyhhuang

JiraIssue: DS-51673

### DIFF
--- a/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/bulk/BulkOperation.java
+++ b/scim2-sdk-common/src/main/java/com/unboundid/scim2/common/bulk/BulkOperation.java
@@ -509,7 +509,7 @@ public abstract class BulkOperation
         final ObjectNode data)
             throws BulkRequestException
     {
-      // Use an object to eliminate potential variance from source JSON data.
+      // Re-encode the source JSON into an object to reformat the input.
       this(path, asPatchRequest(data));
     }
 
@@ -729,7 +729,7 @@ public abstract class BulkOperation
                                     @NotNull final PatchOperation op1,
                                     @Nullable final PatchOperation... ops)
   {
-    return patch(endpoint, toList(op1, ops));
+    return new BulkPatchOperation(endpoint, new PatchRequest(toList(op1, ops)));
   }
 
   /**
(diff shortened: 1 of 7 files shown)


scim2-sdk-common/src/main/java/com/unboundid/scim2/common/bulk/BulkResourceMapper.java
```diff
@@ -142,26 +139,15 @@ public static synchronized <T extends ScimResource> void add(
...
-      schema = SchemaUtils.getSchema(clazz);
-    }
-    catch (IntrospectionException e)
-    {
-      Debug.debugException(e);
-      throw new RuntimeException(e);
-    }
-
-    if (schema == null || schema.getId() == null)
+    Schema schema = clazz.getAnnotation(Schema.class);
```
kqarryzada, line 142: SchemaUtils was doing a lot of extra work that we don't actually need here. To eliminate exception potential and make this simpler, we now fetch the annotation directly.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/bulk/BulkOperation.java
```diff
@@ -77,29 +78,15 @@
...
  *        information, see {@link ETagConfig}.
  *   <li> {@code data}: The JSON body/payload of the SCIM request. For example,
  *        in a POST operation, this represents the resource to create.
  * </ul>
  * <br><br>
  *
- * To construct a bulk operation, use one of the static methods defined on this
- * class. Some examples include:
- * <ul>
- *   <li> {@link #post(String, ScimResource)}
```
kqarryzada, line 92: This list broke the flow of the explanation, and it's effectively re-explained with the example below.

vyhhuang (approved):

dougbulkley (approved):

PR #304, opened 2026-07-10
Fix customization of JsonReadFeatures.

kqarryzada:
This commit fixes an issue where setting a custom Jackson JsonReadFeature property was not reflected by the SCIM SDK's JsonMapper.

Reviewer: dougbulkley
Reviewer: vyhhuang

JiraIssue: DS-51680

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,9 @@ was missing, ensuring that a `PatchRequest` is always available at runtime when 
 
 Corrected an annotation in `PatchRequest` which did not mark `Operations` with a multiValueClass.
 
+Fixed an issue where Jackson `JsonReadFeature` properties were not used despite the property being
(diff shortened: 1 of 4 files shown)


vyhhuang (approved):

dougbulkley (approved):
I don't have any experience with JsonReadFeature class. Wondering why you only test the  ALLOW_SINGLE_QUOTES constant and not the other ones?

kqarryzada:
@dougbulkley, this is really because most features aren't as relevant for filter processing. For example, `JsonReadFeature.ALLOW_TRAILING_COMMA` doesn't work since we explicitly don't allow that in a filter expression anyway. I suppose the numeric handling ones do fit, though it arguably doesn't add substantial coverage and could be harder to maintain if such parameters are deprecated by Jackson in the future.

PR #310, opened 2026-07-31
Update equals() methods.

kqarryzada:
This commit updates the formatting of equals() methods within the SCIM SDK. The primary benefit of this change is that it results in better unit test code coverage statistics. This change does not include new unit tests, since equality evaluation is exercised much better in other Ping Identity test suites.

This also includes small updates to classes that did not consistently invoke the superclass in both equals() and hashCode().

Reviewer: vyhhuang

JiraIssue: DS-51721


scim2-sdk-common/src/main/java/com/unboundid/scim2/common/bulk/BulkOperation.java
```diff
@@ -832,12 +832,9 @@ public boolean equals(@Nullable final Object o)
     {
       return true;
```
kqarryzada, line 833: This could have been folded in with a ternary operator in the return statement, but it wasn't worthwhile in my opinion.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/Path.java
```diff
@@ -254,18 +254,9 @@ public boolean equals(@Nullable final Object o)
...
-      Element element = (Element) o;
-
-      if (!toLowerCase(attribute).equals(toLowerCase(element.attribute)))
-      {
-        return false;
-      }
-      return Objects.equals(valueFilter, element.valueFilter);
+      return o instanceof Element element
+          && Objects.equals(valueFilter, element.valueFilter)
+          && attribute.equalsIgnoreCase(element.attribute);
```
kqarryzada, line 259: I updated this to equalsIgnoreCase() since it's more efficient.

vyhhuang (approved):

PR #313, opened 2026-08-21
Update DateTimeUtils#parse to support date strings

kqarryzada:
Updated the DateTimeUtils class to allow evaluation of date strings without a timestamp, e.g., "1970-01-01". Although this is not generally used in SCIM, it broadens compatibility in certain cases, such as if a date is provided in a SCIM filter without a timestamp. Date strings will be given a timestamp marking the beginning of the day.

This does incur a small performance penalty for identifying invalid date strings. However, this tradeoff is worthwhile to provide a slightly more useful parsing implementation. In general, most invocations are expected to provide a valid string, and that primary path is unaffected.

Reviewer: dougbulkley
Reviewer: vyhhuang

JiraIssue: DS-51859


scim2-sdk-common/src/test/java/com/unboundid/scim2/common/DateTimeUtilsTest.java
```diff
@@ -311,8 +360,6 @@ public Object[][] invalidTimestamps()
   {
     return new Object[][]
         {
-            new String[] { "1989-03-14T17:56:47+09:999" },
-            new String[] { "1989-03-14T17:56:47+99:00" },
```
kqarryzada, line 315: I removed these since timezones don't get the useful "Invalid value for..." message that matches the regex in the test above.

scim2-sdk-common/src/test/java/com/unboundid/scim2/common/DateTimeUtilsTest.java
```diff
@@ -375,12 +447,15 @@ public Object[][] timestampTestCases()
             // Test a non-standard value for a timezone.
             testcase("2079-10-18T08:53:50-11:07", 3464884850000L, zone(-11, -7)),
 
+            // RFC 7643 states that "A date time... has no case sensitivity".
+            testcase("2079-10-18t08:53:50-08:00", 3464873630000L, zone(-8)),
+            testcase("2098-06-17T20:13:43z", 4053874423000L, UTC),
+            testcase("2098-06-17t20:13:43z", 4053874423000L, UTC),
```
kqarryzada, line 453: These were already supported, but now there is a dedicated regression test.

vyhhuang (approved):

dougbulkley (approved):
Claude suggests these additions into the datesWithoutTimestamps data provider, I'll leave it up to you to implement or not:

            // RFC 7643 states that "A date time... has no case sensitivity".
            // This should also apply to the date-only fallback parsing path.
            { "1970-01-01z", 0L, UTC },

            // Non-zero-minute offsets, to match the variety already covered
            // for full timestamps in 'timestampTestCases'.
            { "1970-01-01+05:30", -19800000L, zone(5, 30) },
            { "1970-01-01-09:30", 34200000L, zone(-9, -30) },

            // An explicit "+00:00" offset should be equivalent to "Z".
            { "1970-01-01+00:00", 0L, UTC },

kqarryzada:
> Claude suggests these additions into the datesWithoutTimestamps data provider

Thanks Doug. I don't anticipate a strong need to support these edge cases if we ever update the parser, so I'll forgo this for the time being.

PR #314, opened 2026-09-11
Support Device Schema Extensions (RFC 9944)

kqarryzada:
Added model classes, extensions, and new fields as defined by RFC 9944.
This RFC describes an update to the SCIM standard for managing user IoT
devices, and the bulk of these changes have been placed in a new,
dedicated "devices" package. The DeviceResource class-level Javadoc
contains an introduction to device resources and how to leverage them
with this SDK.

This change includes support in scim2-sdk-server for evaluating device
resource and related object compliance with the SchemaChecker. Note that
explicit client changes were not required, as the DeviceResource class
can leverage the existing client APIs like any ScimResource object.

The release version has been set to 6.1.0-SNAPSHOT.

Reviewer: dougbulkley
Reviewer: vyhhuang

JiraIssue: DS-51992


scim2-sdk-common/src/main/java/com/unboundid/scim2/common/types/devices/EndpointAppDeviceExtension.java
```diff
@@ -0,0 +1,358 @@
...
+    name = "Application Endpoint Extension",
+    description = "Endpoint applications extension for a Device resource")
+public class EndpointAppDeviceExtension extends DeviceExtension
+{
+  // Like $ref fields, this is marked as not required since clients will not set
+  // this value in requests they send to SCIM services.
+  @Nullable
+  @Attribute(description =
+      "The URI of the enterprise device-control endpoint application.",
+      isRequired = false,
```
dougbulkley, line 146: In the RFC this is required. I see your comment explaining the deviation. You might consider noting this deviation in the CHANGELOG.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/types/devices/CertificateInfo.java
```diff
@@ -0,0 +1,184 @@
...
+{
+  @Nullable
+  @Attribute(description = "The certificate authority for the endpoint app.",
+      isRequired = false)
+  private String rootCA;
+
+  @Nullable
+  @Attribute(description =
+      "The subject name for the endpoint application certificate.",
+      isRequired = false)
```
dougbulkley, line 90: The subjectName for CertificateInfo is required and case exact in the RFC tables. But I also found that the RFC is in disagreement with itself because it also states subjectName "...is not required and not case sensitive." An RFC bug?

dougbulkley (approved):

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/types/devices/EndpointAppReference.java
```diff
@@ -0,0 +1,201 @@
...
+      isCaseExact = false)
+  private String value;
+
+  @Nullable
+  @Attribute(description = "The URI of the EndpointApp resource.",
+      isRequired = false,
+      isCaseExact = true,
+      referenceTypes = { "EndpointApp" },
+      mutability = AttributeDefinition.Mutability.READ_ONLY)
+  @JsonProperty("$ref")
```
vyhhuang, line 67: Like with the 'deviceControlEnterpriseEndpoint' field in EndpointAppDeviceExtension, it might be worth explicitly stating that this field is not required in the changelog.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/types/devices/ZigbeeDeviceExtension.java
```diff
@@ -0,0 +1,224 @@
...
+  private final List<String> versionSupport;
+
+  @NotNull
+  @Attribute(description = "The IEEE EUI-64 address of this Zigbee device."
+      + " Must match the pattern: " + EUI64_PATTERN,
+      isRequired = true,
+      isCaseExact = false,
+      mutability = AttributeDefinition.Mutability.READ_WRITE,
+      returned = AttributeDefinition.Returned.DEFAULT,
+      uniqueness = AttributeDefinition.Uniqueness.NONE,
```
vyhhuang, line 135: Nit- both of these fields should have the default of `AttributeDefinition.Uniqueness.NONE` -- is there a reason only one is explicitly marked as such?

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/SchemaUtils.java
```diff
@@ -737,7 +716,7 @@ public static String getSchemaUrn(@NotNull final Class<?> cls)
     // The 'schemaId' is the URN. Make sure it begins with the "urn:" prefix.
     // The 'name' field is a human-friendly name for the object.
     String schemaId =
-        SchemaUtils.getSchemaIdFromAnnotation(cls);
+          SchemaUtils.getSchemaIdFromAnnotation(cls);
```
vyhhuang, line 719: Uber-nit: unnecessary whitespace change.

vyhhuang (approved):
I had a couple of optional suggestions, but these changes look good to me.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/types/devices/EndpointAppDeviceExtension.java
kqarryzada: I've added an entry to the CHANGELOG for this.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/types/devices/CertificateInfo.java
kqarryzada: I set this as not required because of that line in the RFC. Now that you mention this, though, all other references in the document mark it as required, so it seems very likely that the "is not required" statement is a mistake. I did find it strange that it was technically valid to have a JSON object with a present but empty certificateInfo field.

Since there's strong evidence the intent was for this field to be required, I've updated the implementation to reflect that. I also got rid of the helper method that tracked the awkward state of "null or empty certificateInfo object". Thanks for bringing this up.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/types/devices/ZigbeeDeviceExtension.java
kqarryzada: No, I just missed this while cleaning up these references. Thanks for mentioning this.

scim2-sdk-common/src/main/java/com/unboundid/scim2/common/utils/SchemaUtils.java
kqarryzada: This was a result of a bad revert. I've fixed this.

